PR #3, opened 2017-12-04
New generated code

alexander-fenster:
This is a new generated code for pubsub. We need to make sure it works with the existing manual code under src/.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "mocha": "^3.5.3",
     "nyc": "^11.3.0",
     "power-assert": "^1.4.4",
-    "prettier": "^1.8.2",
+    "prettier": "^1.9.1",
     "proxyquire": "^1.7.10"
   }
 }
--- a/package.json
+++ b/package.json
@@ -56,7 +56,6 @@
     "prettier": "repo-tools exec -- prettier --write src/*.js src/*/*.js samples/*.js samples/*/*.js test/*.js test/*/*.js system-test/*.js system-test/*/*.js",
     "docs": "repo-tools exec -- jsdoc -c .jsdoc.js",
     "publish-module": "node ../../scripts/publish.js pubsub",
-    "system-test": "mocha system-test/*.js --no-timeouts --bail",
     "generate-scaffolding": "repo-tools generate all && repo-tools generate lib_samples_readme -l samples/ --config ../.cloud-repo-tools.json"
   },
   "dependencies": {
@@ -65,11 +64,13 @@
     "async-each": "^1.0.1",
     "extend": "^3.0.1",
     "google-auto-auth": "^0.7.1",
-    "google-gax": "^0.13.0",
+    "google-gax": "^0.14.2",
     "google-proto-files": "^0.13.1",
     "grpc": "^1.7.2",
     "is": "^3.0.1",
+    "lodash.merge": "^4.6.0",
     "lodash.snakecase": "^4.1.1",
+    "protobufjs": "^6.8.1",
     "uuid": "^3.1.0"
   },
   "devDependencies": {
--- a/src/index.js
+++ b/src/index.js
@@ -77,7 +77,7 @@ function PubSub(options) {
 
   this.options = extend(
     {
-      scopes: v1.ALL_SCOPES,
+      'grpc.keepalive_time_ms': 300000,
       'grpc.max_receive_message_length': 20000001,
       libName: 'gccl',
       libVersion: PKG.version,
@@ -189,7 +189,7 @@ PubSub.prototype.createSubscription = function(topic, name, options, callback) {
 
   this.request(
     {
-      client: 'subscriberClient',
+      client: 'SubscriberClient',
       method: 'createSubscription',
       reqOpts: reqOpts,
       gaxOpts: options.gaxOpts,
@@ -249,7 +249,7 @@ PubSub.prototype.createTopic = function(name, gaxOpts, callback) {
 
   this.request(
     {
-      client: 'publisherClient',
+      client: 'PublisherClient',
       method: 'createTopic',
       reqOpts: reqOpts,
       gaxOpts: gaxOpts,
@@ -352,7 +352,7 @@ PubSub.prototype.getSnapshots = function(options, callback) {
 
   this.request(
     {
-      client: 'subscriberClient',
+      client: 'SubscriberClient',
       method: 'listSnapshots',
       reqOpts: reqOpts,
       gaxOpts: gaxOpts,
@@ -479,7 +479,7 @@ PubSub.prototype.getSubscriptions = function(options, callback) {
 
   this.request(
     {
-      client: 'subscriberClient',
+      client: 'SubscriberClient',
       method: 'listSubscriptions',
       reqOpts: reqOpts,
       gaxOpts: gaxOpts,
@@ -598,7 +598,7 @@ PubSub.prototype.getTopics = function(options, callback) {
 
   this.request(
     {
-      client: 'publisherClient',
+      client: 'PublisherClient',
       method: 'listTopics',
       reqOpts: reqOpts,
       gaxOpts: gaxOpts,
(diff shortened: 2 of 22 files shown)


package.json
```diff
@@ -47,7 +47,7 @@
     "mkamioner <mkamioner@gmail.com>"
   ],
   "scripts": {
-    "system-test": "repo-tools test run --cmd mocha -- system-test/*.js --no-timeouts",
+    "system-test": "mocha system-test/*.js --no-timeouts --bail",
```
stephenplusplus: I think `repo-tools test run ...` is the right command. That's the one that nodejs-storage uses: https://github.com/googleapis/nodejs-storage/blob/4da861fc7c03fa4b4539966539937f96e66bf757/package.json#L56

src/index.js
```diff
@@ -686,14 +686,48 @@ PubSub.prototype.request = function(config, callback) {
...
+      gaxClient = new v1.PublisherClient(this.options);
+    }
+    else if (config.client == 'subscriberClient') {
+      gaxClient = new v1.SubscriberClient(this.options);
+    }
+    else {
+      throw new Error("Client is unknown: " + config.client);
+    }
+    var scopes = gaxClient.constructor.scopes;
+    extend(this.options, { scopes });
```
stephenplusplus: Might as well just assign `this.options.scopes = gaxClient.constructor.scopes`.

Although, it might be better to do this the way @lukesneeringer did for Logging: https://github.com/googleapis/nodejs-logging/blob/bf3936c7979b4b4d6a55f48604b0c266bc58fc90/src/index.js#L116-L130. It combines all of the scopes from all required classes.

src/index.js
```diff
@@ -686,14 +686,48 @@ PubSub.prototype.request = function(config, callback) {
 
   if (!gaxClient) {
     // Lazily instantiate client.
-    gaxClient = v1(this.options)[config.client](this.options);
+    if (config.client === 'publisherClient') {
```
stephenplusplus: I think it would be better if we went over the places we call `this.request()` and changed `publisherClient` or `subscriberClient` to the correct casing of the class. We used the lowercase to match the class name at the time, but if that's changing, we should change how we refer to it as well.

package.json
alexander-fenster: Agree, not sure why it got edited this way. Reverted.

src/index.js
alexander-fenster: Updated the case to PublisherClient and SubscriberClient.

src/index.js
alexander-fenster: Now getting scopes from all v1 classes.

test/connection-pool.js
```diff
@@ -25,9 +25,6 @@ var proxyquire = require('proxyquire');
 var uuid = require('uuid');
 var util = require('util');
 
-var PKG = require('../package.json');
-var v1 = require('../src/v1');
-
 var fakeUtil = extend({}, common.util);
 var fakeUuid = extend({}, uuid);
 var fakeGrpc = extend({}, grpc);
```
stephenplusplus: We can remove the stuff about `fakeGrpc` now that we aren't using it.

stephenplusplus:
`codecov` doesn't look to be hooked up yet, so I ran `npm test` locally to see the coverage, and it looks like there's a small drop in `index.js` down to 98.33.

test/connection-pool.js
alexander-fenster: Removed

alexander-fenster:
Added several tests to cover new code in src/index.js.

test/index.js
```diff
@@ -1065,6 +1065,102 @@ describe('PubSub', function() {
     });
   });
 
+  describe('request', function() {
```
stephenplusplus: I believe this is meant to read `getClient_`

alexander-fenster: sure :)

src/index.js
```diff
@@ -686,14 +686,59 @@ PubSub.prototype.request = function(config, callback) {
...
+
+    let allScopes = {};
+    for (let clientClass of clientClasses) {
+      for (let scope of clientClass.scopes) {
+        allScopes[scope] = true;
+      }
+    }
+
+    var scopes = Object.keys(allScopes);
+    this.options.scopes = scopes;
```
stephenplusplus: I don't believe we have a test for this functionality (getting the scopes, then assigning them to `option`)

test/index.js
```diff
@@ -1091,6 +1065,114 @@ describe('PubSub', function() {
...
+        pubsub.getClient_(CONFIG, function(err, client2) {
+          assert.strictEqual(null, err);
+          assert.strictEqual(client1, client2);
+          done();
+        });
+      });
+    });
+  });
+
+  describe('request', function() {
```
stephenplusplus: Most of the tests in this block are actually tests for `getClient_`. I'll point out the one that is correct.

test/index.js
```diff
@@ -1091,6 +1065,114 @@ describe('PubSub', function() {
...
+      method: 'fakeMethod',
+      reqOpts: {a: 'a'},
+      gaxOpts: {},
+    };
+
+    beforeEach(function() {
+      pubsub.config = CONFIG;
+    });
+
+    it('should refuse to create unknown client', function(done) {
```
stephenplusplus: This is a test for `getClient_`.

test/index.js
```diff
@@ -1091,6 +1065,114 @@ describe('PubSub', function() {
...
+        callback(error);
+      };
+
+      pubsub.request(CONFIG, function(err) {
+        assert.strictEqual(err, error);
+        done();
+      });
+    });
+
+    it('should replace the project id token on reqOpts', function(done) {
```
stephenplusplus: This is correctly placed. (The project ID token replacement happens in `request`)

test/index.js
```diff
@@ -1091,6 +1065,114 @@ describe('PubSub', function() {
...
+
+    it('should do nothing if sandbox env var is set', function(done) {
+      global.GCLOUD_SANDBOX_ENV = true;
+      pubsub.request(CONFIG, done); // should not fire done
+      global.GCLOUD_SANDBOX_ENV = false;
+      done();
+    });
+  });
+
+  describe('request', function() {
```
stephenplusplus: For `request`, I think we need new tests, and we can re-use one from above. Here's what I would do:

- should call getClient_ correctly (checks that `config` object is received by `getClient_`)
- should return an error from getClient_
- **re-used test**: should replace the project id token on reqOpts
- should call client method correctly (see that it gets the correct `reqOpts`, `config.gaxOpts`, and `callback`)

src/index.js
alexander-fenster: It's verified independently by system tests, but I agree unit test is needed as well. Will add.

test/index.js
alexander-fenster: Sure - there was no getClient_ block before :) I will move them accordingly.

alexander-fenster:
Added more tests for request and moved some tests under getClient_.

stephenplusplus:
@alexander-fenster please check out the latest commits that did some refactoring of the tests.

test/index.js
```diff
@@ -175,6 +177,17 @@ describe('PubSub', function() {
...
 
+    it('should combine all required scopes', function() {
+      v1ClientOverrides.SubscriberClient = util.noop;
+      v1ClientOverrides.SubscriberClient.scopes = ['a', 'b', 'c'];
+
+      v1ClientOverrides.PublisherClient = util.noop;
+      v1ClientOverrides.PublisherClient.scopes = ['a', 'b', 'c', 'd', 'e'];
+
+      var pubsub = new PubSub({});
+      assert.deepEqual(pubsub.options.scopes, ['a', 'b', 'c', 'd', 'e']);
```
alexander-fenster: would be better to add some unique scope to SubscriberClient, otherwise you don't actually check that it's a union of two sets

stephenplusplus: Great point. Would you mind making that change?

PR #7, opened 2017-12-12
refactor before hook to wait for topics

### LINKED ISSUE #5: Remove setTimeout() in system tests
The first two system tests (system-tests/pubsub.js, Topic: should be listed, should list topics in a stream) are flaky, possibly because of the 5 sec timeout in the before() block that waits for the topics to be actually created and, in some cases, might be not enough.

Let's make a small refactor of the test to avoid using setTimeout in the before() block and try to poll the service instead and wait until three topics are there. We can cancel polling and fail if the topics were not created in some unreasonable long amount of time (say, 2 or 3 minutes).

callmehiphop:
Fixes #5 

- [ ] Tests and linter pass

### DIFF
--- a/system-test/pubsub.js
+++ b/system-test/pubsub.js
@@ -92,6 +92,25 @@ describe('pubsub', function() {
     );
   }
 
+  function waitForTopics(done) {
+    async.retry(
+      {
+        times: 6,
+        interval: 10000,
+      },
+      function(callback) {
+        pubsub.getTopics(function(err, topics) {
+          if (!err && !topics.length) {
+            err = new Error('No topics found.');
+          }
+
+          callback(err, topics);
+        });
+      },
+      done
+    );
+  }
+
   before(function(done) {
     // create all needed topics
     async.each(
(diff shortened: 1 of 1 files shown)


system-test/pubsub.js
```diff
@@ -92,6 +92,25 @@ describe('pubsub', function() {
...
 
+  function waitForTopics(done) {
+    async.retry(
+      {
+        times: 6,
+        interval: 10000,
+      },
+      function(callback) {
+        pubsub.getTopics(function(err, topics) {
+          if (!err && !topics.length) {
```
stephenplusplus: I think we should test to make sure the topics that are returned match the topics that we created, so probably checking their names. If a cleanup cycle didn't execute from a previous run, this could be a false positive.

callmehiphop:
@alexander-fenster well that's embarrassing!

system-test/pubsub.js
callmehiphop: PTAL!

PR #37, opened 2018-01-09
allow ack and nack to be destructured

### LINKED ISSUE #32: Message functions don't work with destructuring
On version 0.15 of the pubsub library, destructuring a message makes the `ack` and `nack` functions nonfunctional, because of the different `this` parameter passed to them once destructured.

#### Environment details

  - OS: Debian Jessie 8
  - Node.js version: 8.9.1
  - npm version: 5.5.1
  - @google-cloud/pubsub version: 0.15

#### Steps to reproduce

  1. The following code reproduces the problem. This is typescript, but should be enough to illustrate repro code.
```ts
import * as PubSub from "@google-cloud/pubsub";
import { inspect } from "util";

(async () => {
  let pubsub = PubSub();
  let [topic] = await pubsub.createTopic("testTopic");
  let [sub] = await topic.createSubscription("testSub");
  sub.on("message", async ({ id, ack }) => {
    console.log("testing ack");
    await ack();
  });
  topic.publisher().publish(Buffer.from("Hi!"));
})();
process.on("unhandledRejection", (err) => console.log("err", inspect(err)))
```
  2. The log spits out:
```
err TypeError: Cannot read property 'ackId' of undefined
    at Subscription.breakLease_ (/XXX/node_modules/@google-cloud/pubsub/src/subscription.js:355:59)
    at Subscription.ack_ (/XXX/node_modules/@google-cloud/pubsub/src/subscription.js:289:8)
    at ack (/XXX/node_modules/@google-cloud/pubsub/src/connection-pool.js:306:25)
    at Object.<anonymous> (/XXX/build/test/test-pubsub.js:19:15)
    at Generator.next (<anonymous>)
    at /XXX/build/test/test-pubsub.js:7:71
    at Promise (<anonymous>)
    at __awaiter (/XXX/build/test/test-pubsub.js:3:12)
    at Subscription.sub.on (/XXX/build/test/test-pubsub.js:17:40)
    at emitOne (events.js:115:13)
```
#### Workaround

Don't destructure the messages :(

#### Proposed fix

Changing the `createMessage` function in `connection-pool.js` to be the following fixes the problem, because we capture the reference to the message in the calls to `subscription.ack_` and `subscription.nack_`:

```js
ConnectionPool.prototype.createMessage = function(connectionId, resp) {
  var self = this;

  var pt = resp.message.publishTime;
  var milliseconds = parseInt(pt.nanos, 10) / 1e6;
  var originalDataLength = resp.message.data.length;

  var message = {
    connectionId: connectionId,
    ackId: resp.ackId,
    id: resp.message.messageId,
    attributes: resp.message.attributes,
    publishTime: new Date(parseInt(pt.seconds, 10) * 1000 + milliseconds),
    received: Date.now(),
    data: resp.message.data,
    // using get here to prevent user from overwriting data
    get length() {
      return originalDataLength;
    },
  };
  message.ack = function() {
    self.subscription.ack_(message);
  },
  message.nack = function() {
    self.subscription.nack_(message);
  }
  return message;
};
```

baronfel:
Fixes #32.

- [x] Tests and linter pass


stephenplusplus:
Ping @callmehiphop

baronfel:
I see that googlebot isn't happy, is there something else I need to do?

callmehiphop:
@baronfel unfortunately I think we have to get rid of your unit tests. We still support Node v4 which does not support destructuring. I think our current unit tests should be sufficient enough to ensure this functionality works as expected.

baronfel:
Hah! That's a career first, writing too many tests. Want me to make that change or should you handle it?

callmehiphop:
@baronfel if you don't mind, that would be great! Otherwise I can get to it sometime today :)

baronfel:
your wish is my command.

PR #52, opened 2018-01-29
Adding return of promise on subscription close when no callback is provided

### LINKED ISSUE #50: Resolve subscription.close() once streams have been flushed.
After using the pub/sub client for test, I have found that I wanted to know when the queue was flushed before exiting.  The comment in the code states that a promise is returned when no callback is provided: https://github.com/googleapis/nodejs-pubsub/blob/master/src/subscription.js#L377 but the code does not return a promise.  

#### Environment details

  - OS: OSX
  - Node.js version: 8.6.0
  - npm version: 5.3.0

#### Steps to reproduce

  1. Open a subscription
  2. Close a subscription with no callback.
  3. Execute `.then()` on the promise from the subscription close method.
  
Expected: the callback in the `.then()` is executed.
Actual: Error calling .then on undefined.

I have completed a fix with a test here: https://github.com/STeveShary/nodejs-pubsub/commit/2bf9c33f92bd43f1020f5467c58c48a796348309 .  Your contribution guide doesn't recommend a PR yet, but I can create one quickly.

STeveShary:
Code and test is provided.

Fixes #50

- [X ] Tests and linter pass
- [ X] Code coverage does not decrease (if any source code was changed)
- [ ]X Appropriate docs were updated (if necessary)

### DIFF
--- a/src/subscription.js
+++ b/src/subscription.js
@@ -389,9 +389,14 @@ Subscription.prototype.close = function(callback) {
   clearTimeout(this.leaseTimeoutHandle_);
   this.leaseTimeoutHandle_ = null;
 
-  this.flushQueues_().then(function() {
-    self.closeConnection_(callback);
-  });
+  var flushPromise = this.flushQueues_();
+  if (callback) {
+    flushPromise.then(function() {
+      self.closeConnection_(callback);
+    });
+  } else {
+    return flushPromise;
+  }
 };
 
 /*!
--- a/test/subscription.js
+++ b/test/subscription.js
@@ -595,6 +595,11 @@ describe('Subscription', function() {
 
       subscription.close(done);
     });
+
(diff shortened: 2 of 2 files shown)


src/subscription.js
```diff
@@ -389,9 +389,14 @@ Subscription.prototype.close = function(callback) {
...
-    self.closeConnection_(callback);
-  });
+  var flushPromise = this.flushQueues_();
+  if (callback) {
+    flushPromise.then(function() {
+      self.closeConnection_(callback);
+    });
+  } else {
+    return flushPromise;
+  }
```
ctavan: This should already be ensured by:

https://github.com/googleapis/nodejs-pubsub/blob/ace59b086c9f72fb1e582398f7cf232b1428b6bc/src/subscription.js#L1240-L1247

PR #56, opened 2018-02-02
Use autoCreate in flow control sample.

### LINKED ISSUE #54: Samples do not show best practices
The samples that demonstrate flow control use them in a CreateSubscription RPC. This is not the recommended pattern, as users only have 1 QPS limit. It is preferred that users attempt to subscribe, and user the error handler to create the subscription if they receive NOT_FOUND from the server. Please correct samples to show this, since we have had users running this in production hit their limits when restarting or trying to scale up their jobs.

stephenplusplus:
Fixes #54

### DIFF
--- a/samples/subscriptions.js
+++ b/samples/subscriptions.js
@@ -125,17 +125,21 @@ function createFlowControlledSubscription(
   // const maxBytes = 10000;
 
   const topic = pubsub.topic(topicName);
-  const subscription = topic.subscription(subscriptionName);
+
+  const options = {
+    flowControl: {
+      maxBytes: maxBytes,
+      maxMessages: maxInProgress,
+    },
+  };
+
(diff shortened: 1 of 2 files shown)


samples/subscriptions.js
```diff
@@ -124,11 +124,14 @@ function createFlowControlledSubscription(
...
+
   // Creates a new subscription
   // Note that flow control configurations are not persistent
-  pubsub
-    .topic(topicName)
-    .createSubscription(subscriptionName, {
+  subscription
+    .get({
+      autoCreate: true,
       flowControl: {
```
callmehiphop: These have to be passed in via `topic.subscription(name, { flowControl: ... })`

stephenplusplus: PTAL!

mdietz94:
Great, thanks!

PR #65, opened 2018-02-13
Send acknowledge/modifyAckDeadline requests in chunks

### LINKED ISSUE #62: subscription.close() leads to "Request payload size exceeds the limit"
#### Environment details

  - OS: Kubernetes Engine
  - Node.js version: 9.3.0
  - npm version: –

#### Steps to reproduce

I have a consumer running on Kubernetes Engine which reads messages from a PubSub subscription and writes them to BigQuery.

I have implemented a clean shutdown logic when my pods get terminated (e.g. during a rolling updated). In that shutdown logic I call `subscription.close()`.

Occasionally calling `subscription.close()` leads to the following error being emitted in the `subscription.on('error')` event:

```
Error: Request payload size exceeds the limit: 524288 bytes.
    at /app/node_modules/grpc/src/client.js:554:15
```

If I inspect the error that I receive in the `subscription.on('error')` handler, I can see that the stack trace contains either:

```
    at Subscription.emit (events.js:159:13)
    at /app/node_modules/@google-cloud/pubsub/src/subscription.js:346:10
    at <anonymous>"    
```

or:

```
    at Subscription.emit (events.js:159:13)
    at /app/node_modules/@google-cloud/pubsub/src/subscription.js:869:10
    at <anonymous>
```

I could trace down the problem as follows:

When `subscription.close()` is called, this in turn calls `flushQueues_()`:

https://github.com/googleapis/nodejs-pubsub/blob/d89287ca9e998c051c5bc4fe5d0bc123481f6a04/src/subscription.js#L381-L395

Within `flushQueues_()`, in case there are pending `acks` or `nacks`, `acknowledge_()` or `modifyAckDeadline_()` are being called will all pending `ackIds/nackIds` respectively:

https://github.com/googleapis/nodejs-pubsub/blob/d89287ca9e998c051c5bc4fe5d0bc123481f6a04/src/subscription.js#L631-L647

Both these methods are rather similar. In any case they will simply make a call to the API with all the `ackIds/nackIds` that were passed from `flushQueues_()`.

Here's `acknowledge_()`:

https://github.com/googleapis/nodejs-pubsub/blob/d89287ca9e998c051c5bc4fe5d0bc123481f6a04/src/subscription.js#L318

And here's `modifyAckDeadline_()`:

https://github.com/googleapis/nodejs-pubsub/blob/d89287ca9e998c051c5bc4fe5d0bc123481f6a04/src/subscription.js#L899-L905

Apparently these API-calls are too big, since the corresponding promises reject with an error that gets then emitted and that I then see in my application code:

https://github.com/googleapis/nodejs-pubsub/blob/d89287ca9e998c051c5bc4fe5d0bc123481f6a04/src/subscription.js#L323-L325

https://github.com/googleapis/nodejs-pubsub/blob/d89287ca9e998c051c5bc4fe5d0bc123481f6a04/src/subscription.js#L910-L912

The error message mentions `524288 bytes` which is precisely `512 KiB`.

My suspicion is that the `acks/nacks` must be sent in batches in case they exceed the maximum payload size for `acknowledge/modifyAckDeadline` requests?

Unfortunately the [REST API docs](https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions/acknowledge) don't say anything about limits.

In any case I would expect that `subscription.close()` allows for a clean shutdown, even if there are many pending `acks/nacks`. The current behavior leads to potential duplicates on my end since pending acks are not correctly being flushed: since the sending of the pending acks itself triggers an error and subsequent crash in my application code the will be re-consumed after the restart of my application, even though they have already been processed.

ctavan:
The acknowledge and modifyAckDeadline API requests seem to have a size
limitation: If you send to many ackIds with one request you may get the
following error:

```
Request payload size exceeds the limit: 524288 bytes.
```

In order to prevent this error, make sure to always send those requests
in chunks that are small enough.

Fixes #62.

- [x] Tests and linter pass

### DIFF
--- a/src/subscription.js
+++ b/src/subscription.js
@@ -17,6 +17,7 @@
 'use strict';
 
 var arrify = require('arrify');
+var chunk = require('lodash.chunk');
 var common = require('@google-cloud/common');
 var events = require('events');
 var extend = require('extend');
@@ -29,6 +30,15 @@ var ConnectionPool = require('./connection-pool.js');
 var Histogram = require('./histogram.js');
 var IAM = require('./iam.js');
 var Snapshot = require('./snapshot.js');
+
+/**
+ * @type {number} - The maximum number of ackIds to be sent in acknowledge/modifyAckDeadline
+ *     requests. There is an API limit of 524288 bytes (512KiB) per acknowledge/modifyAckDeadline
+ *     request. ackIds have a maximum size of 164 bytes, so 524288/164 ~= 3197. Accounting for some
+ *     overhead, a maximum of 3000 ackIds per request should be safe.
+ * @private
+ */
+var MAX_ACK_IDS_PER_REQUEST = 3000;
 
 /**
  * A Subscription object will give you access to your Cloud Pub/Sub
@@ -294,33 +304,41 @@ Subscription.prototype.ack_ = function(message) {
  */
 Subscription.prototype.acknowledge_ = function(ackIds, connId) {
   var self = this;
-  var promise;
 
   ackIds = arrify(ackIds);
 
-  if (!this.isConnected_()) {
-    promise = common.util.promisify(this.request).call(this, {
-      client: 'SubscriberClient',
-      method: 'acknowledge',
-      reqOpts: {
-        subscription: this.name,
-        ackIds,
-      },
-    });
-  } else {
-    promise = new Promise(function(resolve, reject) {
+  var promises = chunk(ackIds, MAX_ACK_IDS_PER_REQUEST).map(function(
+    ackIdChunk
+  ) {
+    if (!self.isConnected_()) {
+      return common.util.promisify(self.request).call(self, {
+        client: 'SubscriberClient',
+        method: 'acknowledge',
+        reqOpts: {
+          subscription: self.name,
+          ackIds: ackIdChunk,
+        },
+      });
+    }
+
+    return new Promise(function(resolve, reject) {
       self.connectionPool.acquire(connId, function(err, connection) {
         if (err) {
           reject(err);
           return;
         }
 
-        connection.write({ackIds}, resolve);
+        connection.write(
+          {
+            ackIds: ackIdChunk,
+          },
+          resolve
+        );
       });
     });
-  }
+  });
 
-  return promise.catch(function(err) {
+  return Promise.all(promises).catch(function(err) {
     self.emit('error', err);
   });
 };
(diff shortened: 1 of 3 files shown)


ctavan:
I'm happy to add test-cases to cover the chunking once we agree on an implementation.

callmehiphop:
@ctavan overall I think this LGTM, did this solve your issue? Any thoughts here @stephenplusplus?

stephenplusplus:
Looks okay to me. Did anyone try the system tests?

ctavan:
@callmehiphop @stephenplusplus I have not seen the error reported in #62 since running with this patched code while I was getting the error on the majority of service restarts before: So I'm pretty confident that it does what it's supposed to :).

Also, we've been running this in production for the last 4 days without any issues. Restarts of our services now no longer produce duplicates that we got due to failed acknowledge requests during service shutdowns.

ctavan:
Oh, I've also added 4 test cases for `acknowledge_/modifyAckDeadline_` for the `streaming/non-streaming` cases respectively which cover the chunking behavior.

Would be great if this would land in the next release!

src/subscription.js
```diff
@@ -294,33 +304,41 @@ Subscription.prototype.ack_ = function(message) {
...
+        client: 'SubscriberClient',
+        method: 'acknowledge',
+        reqOpts: {
+          subscription: self.name,
+          ackIds: ackIdChunk,
+        },
       });
-    });
-  }
+    } else {
```
callmehiphop: Small nitpick - if we're returning, we shouldn't need this `else` block anymore.

src/subscription.js
```diff
@@ -874,40 +892,43 @@ Subscription.prototype.listenForEvents_ = function() {
...
+        method: 'modifyAckDeadline',
+        reqOpts: {
+          subscription: self.name,
+          ackDeadlineSeconds: deadline,
+          ackIds: ackIdChunk,
+        },
       });
-    });
-  }
+    } else {
```
callmehiphop: RE: https://github.com/googleapis/nodejs-pubsub/pull/65/files#r168789103

callmehiphop:
@ctavan looks like our lint task is failing - if you would be so kind as to run `npm run prettier`, I think that should fix it!

src/subscription.js
ctavan: Sure, I'm a big fan of early return as well. Wasn't sure if this was the style being used here as well. Changed it!

src/subscription.js
ctavan: Done!

ctavan:
@callmehiphop whoops, I should have checked linting, sorry for the additional loop.

Should be better now.

callmehiphop:
@ctavan no worries! I went ahead and ran all the system tests locally and they're all green, so I think we're good to go. Thanks again for the contribution!

PR #92, opened 2018-03-09
refactor: flow control updates

### LINKED ISSUE #13: possible memory leak?
_From @JoshFerge on November 27, 2017 19:15_

#### Environment details

  - OS:  Debian GNU/Linux 9.2 (stretch
  - Node.js version: v8.9.1
  - npm version: 5.5.1
  - google-cloud-node version: pubsub-0.15.0
```
"@google-cloud/pubsub": {
      "resolved": "https://registry.npmjs.org/@google-cloud/pubsub/-/pubsub-0.15.0.tgz",
        "@google-cloud/common": "0.13.6",
        "google-auto-auth": "0.7.2",
        "google-gax": "0.13.5",
        "google-proto-files": "0.13.1",
```
#### Steps to reproduce
```
'use strict';
let pubsub = require( '@google-cloud/pubsub' )();

let topic = pubsub.topic( 'my-topic' );
let subscription = topic.subscription( 'test', { flowControl: { maxMessages : 20 } } );
let c = 0;
subscription.on( 'error', function( err ) { console.error( err ); } );
// Register a listener for `message` events.
function onMessage( message ) {
    message.ack(); 
    if ( c++ % 1000 === 0 ) {
        printMemoryUsage();
    }
}

function printMemoryUsage() {
    let obj = process.memoryUsage();
    console.log( 'Date: ' + new Date() + ', Memory: ' + Math.ceil(  obj.heapUsed / 1000 / 1000 , 2 ) + 'Mb' );
}

subscription.on( 'message', onMessage );

```
memory usage quickly climbs and eventually runs out. is there something i'm missing here or is there an internal problem? 

(I have thousands of messages in this topic that are being processed, so that is the scale)

_Copied from original issue: GoogleCloudPlatform/google-cloud-node#2757_

### LINKED ISSUE #50: Resolve subscription.close() once streams have been flushed.

### LINKED ISSUE #66: subscription keeps receiving messages after subscription.close() was called
#### Environment details

  - OS: Kubernetes Engine
  - Node.js version: 9.3.0
  - npm version: –

#### Steps to reproduce

During my investigations on #62 I cluttered the library code with `console.log()` statements ([see this branch for reference](https://github.com/googleapis/nodejs-pubsub/compare/master...contentpass:fix-ack-on-subscription-close-debug)) and found an interesting pattern:

After `subscription.close()` is called, the subscription can still receive many new messages, that all need to be `nack`ed. For the investigation [I've placed a debug statement in the `modifyAckDeadline_()` method](https://github.com/googleapis/nodejs-pubsub/compare/master...contentpass:fix-ack-on-subscription-close-debug#diff-3d29c4447546c72118ed5d5cbf38ab8bR954).

Here's the output of this log statement:

```
{
  "message": "modifyAckDeadline_ non-streaming",
  "acknowledgeInstanceId": "230edb80-718a-4b0a-8e9f-083da5da3fd2",
  "deadline": 0,
  "count": 2202,
  "bytes": 359042
}
{
  "message": "modifyAckDeadline_ non-streaming",
  "acknowledgeInstanceId": "3d749b9f-0156-4120-aee5-fa9c1ce455b0",
  "deadline": 0,
  "count": 5068,
  "bytes": 826668
}
{
  "message": "modifyAckDeadline_ non-streaming",
  "acknowledgeInstanceId": "2b2b66b9-723b-4cb6-a654-8ebad6ff298a",
  "deadline": 0,
  "count": 7889,
  "bytes": 1287063
}
{
  "message": "modifyAckDeadline_ non-streaming",
  "acknowledgeInstanceId": "c7b13ac3-5440-438e-9eac-f6bda0f59bc6",
  "deadline": 0,
  "count": 9270,
  "bytes": 1512777
}
{
  "message": "modifyAckDeadline_ non-streaming",
  "acknowledgeInstanceId": "aa3de1e8-55bd-432f-b9b0-8b1c057a9b76",
  "deadline": 0,
  "count": 11806,
  "bytes": 1926374
}
{
  "message": "modifyAckDeadline_ non-streaming",
  "acknowledgeInstanceId": "eed24ea6-2bce-4137-8ad8-c49dc16e528b",
  "deadline": 0,
  "count": 280,
  "bytes": 45585
}
{
  "message": "modifyAckDeadline_ non-streaming",
  "acknowledgeInstanceId": "9c8653a5-2117-4454-9e96-fd154e75bdf0",
  "deadline": 0,
  "count": 2861,
  "bytes": 466398
}
```

Or looking just at the ackId-counts:

```
  "count": 2202,
  "count": 5068,
  "count": 7889,
  "count": 9270,
  "count": 11806,
  "count": 280,
  "count": 2861,
```

As you can see the counts keep on increasing for a while before the eventually drop. Please note that this all happens _after_ the `subscription.close()` call.

Would it somehow be possible to stop receiving new messages as soon as `.close()` has been called? Or is the observed behavior intended?

### LINKED ISSUE #83: Lease message do no work for long running jobs
Hi, 
#### Environment details

  - OS: Ubuntu 16.04
  - Node.js version: 8.9.4
  - npm version: 5.6.0
  - @google-cloud/pubsub version: 0.16.4

In [that issue](https://github.com/googleapis/nodejs-pubsub/issues/81) @callmehiphop mentioned the following:
> Even if we allowed the deadline to be set, the current code would overwrite it and your message would not get redelivered until your subscription object was closed.

But the message **get redelivered** even when `subscription.close` and `message.ack` are not called.

#### Steps to reproduce
  1. Create topic and subscription
  2. Subscribe to message event
  3. Publish message
  4. Do not call `ack` or `nack` in message handler. Count the number that message handler was called during at least 120 seconds.

```javascript
it('should not redeliver message until subscription is closed', function (done) {
    subscription.on('error', done);

    const onMessage = sinon.spy();
    subscription.on('message', onMessage);

    topic
        .publisher()
        .publish(Buffer.from(uuid()))
        .catch(done);

    setTimeout(() => {
        try {
            sinon.assert.calledOnce(onMessage);
            done();
        } catch (error) {
            done(error);
        }
    }, 120000);
});
```
If you want to run this test you could clone [this repository](https://github.com/muryginm/google-cloud-pubsub-issues).

I almost always have the broken test. Does It mean that lease message does not work?

### LINKED ISSUE #88: Receiving lots of duplicate messages
Hi

I've been getting a lot of duplicate messages recently, specifically when running with a largeish backlog of messages, limited `maxMessages` and delayed acking. I have downgraded to v0.13 (i.e. before StreamingPull) and the dupes have disappeared.

This could possibly be related to https://github.com/googleapis/nodejs-pubsub/issues/68, https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2465 and https://github.com/GoogleCloudPlatform/google-cloud-go/issues/778

#### Environment details

  - OS: ubuntu linux (running under GKE)
  - Node.js version: 9.4.0
  - @google-cloud/pubsub version: 0.16.5
  - subscription ack deadline: 600s

#### Steps to reproduce
The following reproduces the issue fairly reliably in a container on GKE, though the number of duplicates we were seeing in our actual usage was higher than this seems to produce.

**publisher**
```javascript
const NOTIFICATION_TOPIC = 'test';
const pubsub = require('@google-cloud/pubsub')()
const PQueue = require('p-queue');

const publisher = pubsub.topic(NOTIFICATION_TOPIC).publisher();

// The queue ensures there are never more than 500 publishes in flight at
// any one time
const queue = new PQueue({ concurrency: 500 });

function publish() {
  return publisher.publish(Buffer.from('test'));
}

const proms = [];
for (let ii = 0; ii < 2000; ii++) {
  proms.push(queue.add(publish));
}

Promise.all(proms).then(() => console.log('done')).catch(err => console.error(err));
```

**subscriber**
```javascript
const NOTIFICATION_TOPIC = 'test';
const NOTIFICATION_SUBSCRIPTION = 'test-test';

const pubsub = require('@google-cloud/pubsub')();
const ackHistory = new Map();

const topic = pubsub.topic(NOTIFICATION_TOPIC);
const subscription = topic.subscription(NOTIFICATION_SUBSCRIPTION, {
  flowControl: {
    maxMessages: 25
  }
});

var count = 0;
subscription.on('message', message => {
    var ackCount = ackHistory.get(message.ackId) || 0;
    ackCount++;
    ackHistory.set(message.ackId, ackCount);
    if (ackCount > 1) {
      console.log('Already seen this ackId',  message.ackId, ackCount);
    }

    // In our code we went and called some internal services. Simulating this
    // by using a timeout
    setTimeout(() => {
      message.ack();
      console.log('ACKED', count++);
    }, 300);
});
```

Let me know if I can provide any more information

callmehiphop:
Depends on google-gax release including https://github.com/googleapis/gax-nodejs/pull/197

Fixes #13 
Fixes #50 
Fixes #66 
Closes #94 

- [x] Tests and linter pass

### Summary of changes

The overall goal of this PR is to refactor flow control.

Initially we thought a memory leak might be present (#13) however after investigating we found that writing to the server is significantly slower than reading from it and we needed to update our inventory after the write completes. Currently we update our inventory right before we write to the server, assuming that operation will resolve quickly, which sometimes it does not. This caused an increase in memory usage because we were flooding our write buffer with requests faster than they could be processed.

We also discovered a GAX issue, which would report a false positive on the completion time of writing data, this made it virtually impossible to implement any kind of flow control.

By addressing both of these issues, memory usage was cut down significantly and allows users to ack/nack messages as quickly as they please without any repercussions. It will also inadvertently fix #50 since we are now able to monitor when data is finished being written to the server.

Several smaller updates were also made in an attempt to cut back on duplicate messages, however until https://github.com/grpc/grpc-node/issues/123 is resolved, its hard to say for certain if this will fix that issue entirely.

### DIFF
--- a/src/connection-pool.js
+++ b/src/connection-pool.js
@@ -197,12 +197,14 @@ ConnectionPool.prototype.createConnection = function() {
 
     var requestStream = client.streamingPull();
 
-    var readStream = requestStream.pipe(through.obj(function(chunk, enc, next) {
-      chunk.receivedMessages.forEach(function(message) {
-        readStream.push(message);
-      });
-      next();
-    }));
+    var readStream = requestStream.pipe(
+      through.obj(function(chunk, enc, next) {
+        chunk.receivedMessages.forEach(function(message) {
+          readStream.push(message);
+        });
+        next();
+      })
+    );
 
     var connection = duplexify(requestStream, readStream, {objectMode: true});
     var id = uuid.v4();
@@ -218,8 +220,7 @@ ConnectionPool.prototype.createConnection = function() {
       .once(CHANNEL_ERROR_EVENT, onChannelError)
       .once(CHANNEL_READY_EVENT, onChannelReady);
 
-    requestStream
-      .on('status', onConnectionStatus);
+    requestStream.on('status', onConnectionStatus);
 
     connection
       .on('error', onConnectionError)
--- a/src/connection-pool.js
+++ b/src/connection-pool.js
@@ -18,9 +18,11 @@
 
 var arrify = require('arrify');
 var common = require('@google-cloud/common');
+var duplexify = require('duplexify');
 var each = require('async-each');
 var events = require('events');
 var is = require('is');
+var through = require('through2');
 var util = require('util');
 var uuid = require('uuid');
 
@@ -142,13 +144,10 @@ ConnectionPool.prototype.acquire = function(id, callback) {
  * @param {?error} callback.error An error returned while closing the pool.
  */
 ConnectionPool.prototype.close = function(callback) {
+  var self = this;
   var connections = Array.from(this.connections.values());
 
   callback = callback || common.util.noop;
-
-  if (this.client) {
-    this.client.close();
-  }
 
   this.connections.clear();
   this.queue.forEach(clearTimeout);
@@ -167,9 +166,20 @@ ConnectionPool.prototype.close = function(callback) {
   each(
     connections,
     function(connection, onEndCallback) {
-      connection.end(onEndCallback);
+      connection.pause();
+      connection.end(function(err) {
+        connection.cancel();
+        onEndCallback(err);
+      });
     },
-    callback
+    function(err) {
+      if (self.client) {
+        self.client.close();
+        self.client = null;
+      }
+
+      callback(err);
+    }
   );
 };
 
@@ -188,9 +198,20 @@ ConnectionPool.prototype.createConnection = function() {
       return;
     }
 
+    var requestStream = client.streamingPull();
+
+    var readStream = requestStream.pipe(through.obj(function(chunk, enc, next) {
+      chunk.receivedMessages.forEach(function(message) {
+        readStream.push(message);
+      });
+      next();
+    }));
+
+    var connection = duplexify(requestStream, readStream, {objectMode: true});
     var id = uuid.v4();
-    var connection = client.streamingPull();
     var errorImmediateHandle;
+
+    connection.cancel = requestStream.cancel.bind(requestStream);
 
     if (self.isPaused) {
       connection.pause();
@@ -200,10 +221,12 @@ ConnectionPool.prototype.createConnection = function() {
       .once(CHANNEL_ERROR_EVENT, onChannelError)
       .once(CHANNEL_READY_EVENT, onChannelReady);
 
+    requestStream
+      .on('status', onConnectionStatus);
+
     connection
       .on('error', onConnectionError)
       .on('data', onConnectionData)
-      .on('status', onConnectionStatus)
       .write({
         subscription: common.util.replaceProjectIdToken(
           self.subscription.name,
@@ -217,7 +240,7 @@ ConnectionPool.prototype.createConnection = function() {
     function onChannelError() {
       self.removeListener(CHANNEL_READY_EVENT, onChannelReady);
 
-      connection.cancel();
+      requestStream.cancel();
     }
 
     function onChannelReady() {
@@ -241,10 +264,8 @@ ConnectionPool.prototype.createConnection = function() {
       errorImmediateHandle = setImmediate(self.emit.bind(self), 'error', err);
     }
 
-    function onConnectionData(data) {
-      arrify(data.receivedMessages).forEach(function(message) {
-        self.emit('message', self.createMessage(id, message));
-      });
+    function onConnectionData(message) {
+      self.emit('message', self.createMessage(id, message));
     }
 
     function onConnectionStatus(status) {
--- a/src/connection-pool.js
+++ b/src/connection-pool.js
@@ -16,7 +16,6 @@
 
 'use strict';
 
-var arrify = require('arrify');
 var common = require('@google-cloud/common');
 var duplexify = require('duplexify');
 var each = require('async-each');
--- a/src/connection-pool.js
+++ b/src/connection-pool.js
@@ -219,7 +219,9 @@ ConnectionPool.prototype.createConnection = function() {
       .once(CHANNEL_ERROR_EVENT, onChannelError)
       .once(CHANNEL_READY_EVENT, onChannelReady);
 
-    requestStream.on('status', onConnectionStatus);
+    requestStream.on('status', function(status) {
+      setImmediate(() => onConnectionStatus(status));
+    });
 
     connection
       .on('error', onConnectionError)
--- a/src/connection-pool.js
+++ b/src/connection-pool.js
@@ -165,8 +165,10 @@ ConnectionPool.prototype.close = function(callback) {
   each(
     connections,
     function(connection, onEndCallback) {
-      connection.cancel();
-      connection.end(onEndCallback);
+      connection.end(function(err) {
+        connection.cancel();
+        onEndCallback(err);
+      });
     },
     function(err) {
       if (self.client) {
@@ -207,7 +209,7 @@ ConnectionPool.prototype.createConnection = function() {
 
     var connection = duplexify(requestStream, readStream, {objectMode: true});
     var id = uuid.v4();
-    var errorImmediateHandle;
+    var lastStatus;
 
     connection.cancel = requestStream.cancel.bind(requestStream);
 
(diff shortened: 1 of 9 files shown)


src/connection-pool.js
```diff
@@ -167,9 +166,20 @@ ConnectionPool.prototype.close = function(callback) {
   each(
     connections,
     function(connection, onEndCallback) {
-      connection.end(onEndCallback);
+      connection.pause();
+      connection.end(function(err) {
+        connection.cancel();
```
callmehiphop: Calling cancel after we end the stream will resolve #66

src/subscription.js
```diff
@@ -1040,16 +1046,11 @@ Subscription.prototype.openConnection_ = function() {
...
-      return;
-    }
+    self.emit('message', self.leaseMessage_(message));
 
-    if (!pool.isPaused) {
+    if (!pool.isPaused && self.hasMaxMessages_()) {
       pool.pause();
     }
-
-    message.nack();
```
callmehiphop: Per the PubSub team nacking messages like this can lead to an increased number of duplicate messages. Since this PRs main focus is on refactoring flow control, we can remove this line and will hopefully see fewer duplicates.

src/v1/subscriber_client_config.json
```diff
@@ -86,7 +86,7 @@
           "retry_params_name": "messaging"
         },
         "StreamingPull": {
-          "timeout_millis": 60000,
```
callmehiphop: @alexander-fenster I think this value is generated, so anytime we update gapics it gets reverted back to 1 minute. I know it seems like a reasonable amount, by when applying grpc deadlines to streams, it actually kills the stream after the designated time regardless of whether or not a connection was established.

That being said, I bumped it to 15 minutes (I think that's where it was before. Is there something on the gapic side of things that could be changed so we don't have to update this value?

alexander-fenster: Yes, the generator takes it from here https://github.com/googleapis/googleapis/blob/b92c84b696a22dbecbcf53c760bc9195af4406b4/google/pubsub/v1/pubsub_gapic.yaml#L259 and this file is replicated from internal source control. I will make a request to set it to 900000 - is that what you need?

alexander-fenster: It's now updated to 900000. https://github.com/googleapis/googleapis/blob/d90956b8adf6c820969457d848db833328076879/google/pubsub/v1/pubsub_gapic.yaml#L259

callmehiphop:
@stephenplusplus @alexander-fenster this should be ready for a final review, if you would care to take a look!

src/connection-pool.js
```diff
@@ -200,10 +219,13 @@ ConnectionPool.prototype.createConnection = function() {
       .once(CHANNEL_ERROR_EVENT, onChannelError)
       .once(CHANNEL_READY_EVENT, onChannelReady);
 
+    requestStream.on('status', function(status) {
+      setImmediate(() => onConnectionStatus(status));
```
stephenplusplus: Can you add a code comment to explain why we need this?

Also, I think `setImmediate(onConnectionStatus, status)` works.

src/v1/subscriber_client_config.json
tedj: @callmehiphop is there there a way to pass this in options instead of getting it from the internal config only.

PR #108, opened 2018-03-26
Fix publish exmaple, which was handling the response as though it mad…

kamalaboulhosn:
…e the raw underlying publish call instead of the client library call.

- [x] Tests and linter pass


callmehiphop:
Thanks for the fix @kamalaboulhosn

PR #107, opened 2018-03-26
change maxMessages default to 100

callmehiphop:
Relates to https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2810

- [x] Tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)

PR #112, opened 2018-03-28
chore: prep release for 0.18.0

callmehiphop:
## Features

- (#22, #110) A Publisher object now supports configuration of API request parameters. For example, this allows you to specify the number of retries for a request.

## Fixes

- (#109) Muffle extraneous `streamingPull` errors.
- (https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2810, #107) Change Subscription `maxMessages` default value from `Infinity` to `100` to prevent intense memory usage.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@google-cloud/pubsub",
-  "version": "0.17.0",
+  "version": "0.18.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
@@ -87,7 +87,7 @@
       "requires": {
         "array-uniq": "1.0.3",
         "arrify": "1.0.1",
-        "concat-stream": "1.6.1",
+        "concat-stream": "1.6.2",
         "create-error-class": "3.0.2",
         "duplexify": "3.5.4",
         "ent": "2.2.0",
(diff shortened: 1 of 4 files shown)


stephenplusplus:
Made some small changes to the notes.

- Move Features above Fixes to document in same priority of semver
- Remove docs-related fixes. The reader of the release notes probably most cares about code stuff
- Re-worded "gax configuration" to "API request parameters", as well as a little more from that line. `gaxOpts` as an option name throughout the API doesn't hide "gax", so that the user can differentiate between the options we take and what comes from "gax" (a term they likely haven't heard before, but are linked to in the docs to go learn about). In the case of release notes, they should sell what the feature is, so I just dumbed down the lingo, and said what it really is.

Here are the originals, if you think I went too far on any of those points!:

## Fixes

- (#109) Muffle extraneous `streamingPull` errors
- (https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2810, #107) Change Subscription `maxMessages` default value from `Infinity` to `100` to prevent intense memory usage.
- (#108) Fix incorrect Publisher sample.

## Features

- (#22, #110) Add gax configuration support to Publisher object to allow for user specified retries, etc.

PR #113, opened 2018-03-28
Added test for message redelivering

muryginm:
Test for #83 

The aim of this PR is to show the issue with lease message.

I didn't try to integrate new test into existed 'Subscription' scope since it is not going to be merged.

Unfortunately, the issue is not stable, so you need to run test a couple of times to get the issue.


muryginm:
I signed it!

JustinBeckwith:
@stephenplusplus @callmehiphop would y'all mind taking a look at this?

callmehiphop:
@JustinBeckwith I think we can close this, I don't think this was ever meant to be merged in, but instead used a reproduction case for messages being redelivered. My understanding after reviewing some of the other clients is that there is no guarantee that a message won't be redelivered. I think there's still room for improvement in the Node client, but I think that's currently being held up by a [grpc feature request](https://github.com/grpc/grpc-node/issues/123).

PR #116, opened 2018-04-04
docs(subscription): add mention of auto-leasing

callmehiphop:
Relates to #115

- [x] Tests and linter pass


toots:
Thanks guys!

PR #138, opened 2018-06-04
refactor(subscription): use unary rpcs for ack/modack requests

callmehiphop:
- [ ] Tests and linter pass

### DIFF
--- a/src/subscription.js
+++ b/src/subscription.js
@@ -75,6 +75,9 @@ var MAX_ACK_IDS_PER_REQUEST = 3000;
  * @param {string} name The name of the subscription.
  * @param {object} [options] See a
  *     [Subscription resource](https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions)
+ * @param {object} [options.batching] Batching configurations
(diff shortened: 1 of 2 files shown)


callmehiphop:
@stephenplusplus I've pushed several commits, one of which moves all the message related functionality to a separate class, this wasn't really necessary here, but it was something I've been meaning to do for a while and I think will help with developing a couple upcoming PubSub features.

src/subscription.js
```diff
@@ -185,6 +188,13 @@ function Subscription(pubsub, name, options) {
     options.flowControl
   );
 
+  this.batching = extend(
+    {
+      maxMilliseconds: 100,
```
callmehiphop: @kir-titievsky does the PubSub team have any recommendations and/or opinions on how long the client should buffer acks for?

callmehiphop:
@theacodes I must have misunderstood the request, but I thought that the request was to not send the requests over the stream anymore.

theacodes:
Indeed that's the case, but it is still useful to have it be configurable. It's up to you.

PR #147, opened 2018-06-25
refactor(subscription): use unary rpcs for ack/modAck

callmehiphop:
Closes #138 

- [x] Tests and linter pass

### DIFF
--- a/src/subscription.js
+++ b/src/subscription.js
@@ -60,6 +60,10 @@ var Subscriber = require('./subscriber.js');
  * @param {string} name The name of the subscription.
  * @param {object} [options] See a
  *     [Subscription resource](https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions)
+ * @param {object} [options.batching] Batch configurations for sending out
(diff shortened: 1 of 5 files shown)


src/subscription.js
```diff
@@ -60,6 +60,10 @@ var Subscriber = require('./subscriber.js');
  * @param {string} name The name of the subscription.
  * @param {object} [options] See a
  *     [Subscription resource](https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions)
+ * @param {object} [options.batching] Batch configurations for sending out
+ *     Acknowledge and ModifyAckDeadline requests.
+ * @param {number} [options.batching.maxMilliseconds] The maximum amount of time
```
stephenplusplus: I think we can put that default value like this: `[options.batching.maxMilliseconds=100]`.

stephenplusplus (approved):

src/subscription.js
callmehiphop: That was my initial thought too, but it appears we've been documenting default values the old way in PubSub so I went this way for the sake of consistency.

PR #211, opened 2018-08-10
chore: convert to es classes

JustinBeckwith:
This is a mostly automated change with VS Code that converts us to es6 classes.  All system tests pass 👍


jkwlui (approved):
magical!

PR #218, opened 2018-08-23
test: fix import sample failures

JustinBeckwith:


stephenplusplus:
I think this will fix #219!

JustinBeckwith:
Pretty sure it won't :) I was running into some weird `replaceProjectIdToken` bugs even with that fix.  @crwilcox I remember you mentioning these elsewhere as well - am I misremembering?

stephenplusplus:
Can we merge this and see what we get? Is Kokoro failing a non-merger?

JustinBeckwith:
Yeah, sorry, the kokoro things can be safely ignored right now.

PR #226, opened 2018-08-28
fix: Return error if subscription name already exists.

### LINKED ISSUE #122: Duplicate subscription names silently fail
#### Environment details

  - OS: OSX 10.13.3
  - Node.js version: v8.9.1
  - npm version: 5.6.0
  - `@google-cloud/pubsub` version: 0.18.0

#### Steps to reproduce

1. Create topic `foo`
2. Create topic `bar`
3. Create subscription `quix` on topic `foo`
4. Create subscription `quix` on topic `bar` (this is successful?) 

If a subscription name already exists for a topic the client doesn't ensure it matches the same topic for the existing resource. 

This check allows the subscription to appear to be created successfully despite <topic, name> pair not matching the original resource:
https://github.com/googleapis/nodejs-pubsub/blob/master/src/index.js#L249

Error caught internally:
```
err is:  { Error: Resource already exists in the project (resource=quix).
    at /tmp/test/grpc/src/client.js:554:15
  code: 6,
  metadata: Metadata { _internal_repr: {} },
  note: 'Exception occurred in retry method that was not classified as transient' }
```

Ideally `createSubscription` would fail when the topic name doesn't match the existing resource.

stephenplusplus:
Fixes #122 

This could be a breaking change, depending on if you consider this a bug that we fixed or a behavior that we switched.

Before:
```js
topic.createSubscription('my-subscription', function(err, subscription) {})

// in the future...
topic.createSubscription('my-subscription', function(err, subscription) {})
```

After:
```js
topic.createSubscription('my-subscription', function(err, subscription) {})

// in the future...
topic.createSubscription('my-subscription', function(err, subscription) {
  // New error returned saying you already have a subscription with this name!
})
```

Here's a whole talk about some of the history: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/1257


JustinBeckwith:
@kinwa91 lets get this in 0.20 too.

stephenplusplus:
You can do:

```js
topic.createSubscription('my-subscription', function(err, subscription) {})

// in the future...
const subscription = topic.subscription('my-subscription')

subscription.get({ autoCreate: true }, function(err) {
  // No more `ALREADY_EXISTS` error!
})
```

Alternatively, if you know the subscription exists, you can just start using it right away:

```js
const subscription = topic.subscription('my-subscription')

subscription.on('message', msg => {})
```

PR #235, opened 2018-08-31
fix: (sample tests): Validate short names.

stephenplusplus:


samples/.eslintrc.yml
```diff
@@ -1,3 +1,4 @@
 ---
 rules:
   no-console: off
+  no-unused-vars: off
```
stephenplusplus: @JustinBeckwith this allows us to create variables in the samples that don't require using. In this case, it helps the users see how to get a reference to the `topics` or `subscriptions` we return. We don't need to use the variables, though.

So, I allowed it here (	samples/`), and disallowed it in `samples/system-test`.

samples/subscriptions.js
```diff
@@ -95,7 +95,7 @@ function createSubscription(topicName, subscriptionName) {
     .createSubscription(subscriptionName)
     .then(results => {
       const subscription = results[0];
```
stephenplusplus: Example of unused variable.

PR #242, opened 2018-09-05
Remove call to unref()

marcusnielsen:

- [ ] Tests and linter pass


marcusnielsen:
Issue was with our test runner `jest`, and not pubsub.

Issue #280, opened 2018-10-04
Publisher attempts pushing empty payload when first message published larger than maxBytes


Legogris:
Even if all our messages seemed to be publishing successfully and our logs looked fine, but every now and then we were seeing instances of `INVALID_ARGUMENT` for publish requests Stack Driver.
After a lot of digging it turned out that if a single message hits `maxBytes`, the publisher will attempt to publish queued messages, even if there are no queued messages.

This was made worse by the fact that when using the promise version of `publish`, errors in sending of batched messages are silenced (this is another issue that would be great to have addressed, perhaps by allowing attaching an event handler on the publisher).

I have a PR ready for this.

#### Environment details
  - OS: NixOS / ubuntu
  - Node.js version: v8.11.3
  - npm version: 5.6.0
  - `@google-cloud/pubsub` version: 0.20.1

#### Steps to reproduce
  1. Instantiate a publisher on a topic
  2. Publish single message that is larger than `maxBytes` when no messages are already queued.
  3. Publisher first attempts publishing empty payload before actually publishing message
  4. PubSub API responds with `HTTP 400` but this error is silenced.

callmehiphop:
Closing per #281. Thanks again @Legogris!

PR #281, opened 2018-10-04
Fix maxBytes batching sending empty messages

### LINKED ISSUE #279: Publisher not pushing immediately when single message is larger than maxBytes
Publisher immediately publishes a message that is exactly equal to `maxBytes`. Is there any particular reason why this is only done when there is an exact match (and then queued for sending later any way) rather than when it's bigger or equal?

I have a PR ready for this.

#### Environment details
  - OS: NixOS / ubuntu
  - Node.js version: v8.11.3
  - npm version: 5.6.0
  - `@google-cloud/pubsub` version: 0.20.1

#### Expected behavior
  1. Post single message larger than maxBytes
  2. Message gets published immediately 

#### Actual behavior
  1. Publish single message larger than maxBytes
  2. Message is queued

Legogris:
Fixes #279 and #280.
Regression test added.
Please let me know if you prefer me to split this up in separate PRs.

- [x] Tests and linter pass
- [x] Code coverage does not decrease
- [x] Appropriate docs were updated (not necessary)

### DIFF
--- a/src/publisher.js
+++ b/src/publisher.js
@@ -163,7 +163,10 @@ class Publisher {
     const opts = this.settings.batching;
     // if this message puts us over the maxBytes option, then let's ship
     // what we have and add it to the next batch
-    if (this.inventory_.bytes > 0 && this.inventory_.bytes + data.length > opts.maxBytes) {
+    if (
+      this.inventory_.bytes > 0 &&
+      this.inventory_.bytes + data.length > opts.maxBytes
+    ) {
       this.publish_();
     }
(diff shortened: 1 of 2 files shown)


JustinBeckwith:
Thanks for the PR @Legogris!  If you can go through the CLA signing thingy, we'll take a look!

Legogris:
I signed it!

callmehiphop:
@JustinBeckwith looks like merging is blocked because of the CLA, are there any other steps we need to take?

PR #305, opened 2018-10-30
Samples to use async/await

muraliQlogic:
Works towards googleapis/google-cloud-node#2865 and googleapis/google-cloud-node#2869 (it's a good idea to open an issue first for discussion)
- [ ] Tests and linter pass


samples/subscriptions.js
```diff
@@ -363,75 +307,66 @@ function synchronousPull(projectName, subscriptionName) {
...
+    messages[message.ackId] = [message.message.data, false];
+    worker(message);
+  });
+
+  let numProcessed = 0;
+
+  // setInterval() gets run every 10s.
+  const interval = setInterval(function() {
+    // Every 10s, we do a check on the processing states of the messages.
+    Object.keys(messages).forEach(async ackId => {
```
JustinBeckwith: With the change to make this async, it likely becomes a `.map` and `await Promise.all`

samples/subscriptions.js
```diff
@@ -468,49 +403,48 @@ function listenForOrderedMessages(subscriptionName, timeout) {
   };
 
   // Listen for new messages until timeout is hit
-  return new Promise(resolve => {
+  await new Promise(resolve => {
```
JustinBeckwith: We should switch this over to a full async function, and use the `await new Promise(r => setTimeout(r, xxx))` pattern

samples/subscriptions.js
```diff
@@ -468,49 +403,48 @@ function listenForOrderedMessages(subscriptionName, timeout) {
...
-      } else {
-        // Have not yet processed the message on which this message is dependent
-        return false;
-      }
-    });
+  });
+  // Pub/Sub messages are unordered, so here we manually order messages by
+  // their "counterId" attribute which was set when they were published.
+  const outstandingIds = Object.keys(outstandingMessages).map(counterId =>
+    parseInt(counterId, 10)
```
JustinBeckwith: Use `Number` instead of `parseInt`

samples/subscriptions.js
```diff
@@ -567,20 +501,13 @@ function getSubscriptionPolicy(subscriptionName) {
...
-    .iam.getPolicy()
-    .then(results => {
-      const policy = results[0];
-      console.log(`Policy for subscription: %j.`, policy.bindings);
-    })
-    .catch(err => {
-      console.error('ERROR:', err);
-    });
+  const [policy] = await pubsub.subscription(subscriptionName).iam.getPolicy();
+  console.log(`Policy for subscription: %j.`, policy.bindings);
```
JustinBeckwith: just use basic string interpolation here

JustinBeckwith:
👋 This has a lot of merge conflicts now :/ Apologies!

PR #329, opened 2018-11-11
refactor: drop dependencies on google-proto-files and async

JustinBeckwith:


system-test/pubsub.ts
```diff
@@ -251,52 +207,24 @@ describe('pubsub', function() {
...
       // Delete subscriptions
-      async.each(
-        SUBSCRIPTIONS,
-        function(sub, callback) {
-          sub.delete(callback);
-        },
-        function(err) {
-          assert.ifError(err);
-          topic.delete(done);
+      return SUBSCRIPTIONS.map(async s => {
```
kinwa91: This is running asynchronous - should use an async function - 
```javascript
after(async () =>
  Promise.all(
    // Delete subscriptions
    SUBSCRIPTIONS.map(async s => {
    ...
```

PR #356, opened 2018-11-21
improve types (2)

kinwa91:


src/iam.ts
```diff
@@ -32,7 +30,7 @@ import { CallOptions } from 'google-gax';
  * @param {object} apiResponse The full API response.
  */
 export interface GetPolicyCallback {
-  (err?: Error|null, acl?: Policy, apiResponse?: r.Response): void;
```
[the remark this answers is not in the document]
kinwa91: I wasn't sure `r.Response` is what's returned with gRPC calls - we don't have the `request` dependency, so not sure why `tsc` didn't complain (perhaps it shows up as a transitive dependency?)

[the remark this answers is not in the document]
JustinBeckwith: Can you do a console.log and see what actually comes back?  🤣

JustinBeckwith:
@kinwa91 I am so sorry for what happened here.

src/iam.ts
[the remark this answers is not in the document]
kinwa91: Ha! there's no `apiResponse`. The jsdoc comment lied! 👿

src/iam.ts
```diff
@@ -23,49 +23,47 @@ import * as arrify from 'arrify';
...
-export type SetPolicyResponse = [Policy, r.Response];
+export type SetPolicyResponse = [Policy, object];
 
 /**
  * @typedef {array} GetPolicyResponse
  * @property {object} 0 The policy.
  * @property {object} 1 The full API response.
  */
-export type GetPolicyResponse = [Policy, r.Response];
+export type GetPolicyResponse = [Policy, object];
```
JustinBeckwith: Still curious what all of these `object`(s) are :)

src/iam.ts
```diff
@@ -74,7 +72,8 @@ export type TestIamPermissionsResponse = [object[], r.Response];
  * @param {object} apiResponse The full API response.
  */
 export interface TestIamPermissionsCallback {
-  (err?: Error|null, permissions?: object|null, apiResponse?: r.Response): void;
+  (err?: Error|null, permissions?: {[key: string]: boolean}|null,
```
JustinBeckwith: The `{[key: string]: boolean}` type feels like something we could export and re-use, no?

src/index.ts
```diff
@@ -39,6 +40,81 @@ const {grpc} = new gax.GrpcClient(opts);
  */
 const PROJECT_ID_PLACEHOLDER = '{{projectId}}';
 
+// tslint:disable-next-line:no-any
+export type Metadata = any;
```
JustinBeckwith: Just use `Metadata` from `nodejs-common`

src/index.ts
```diff
@@ -39,6 +40,81 @@ const {grpc} = new gax.GrpcClient(opts);
...
+export type CreateTopicResponse = [Topic, object];
+
+/**
+ * @callback CreateTopicCallback
+ * @param {?Error} err Request error, if any.
+ * @param {Topic} topic The new {@link Topic}.
+ * @param {object} apiResponse The full API response.
+ */
+export interface CreateTopicCallback {
+  (err?: Error|null, topic?: Topic|null, apiResponse?: object): void;
```
JustinBeckwith: Now you have me gun shy, are there apiResponse objs here?

src/index.ts
```diff
@@ -39,6 +40,81 @@ const {grpc} = new gax.GrpcClient(opts);
...
+}
+
+export type Client = 'PublisherClient'|'SubscriberClient';
+
+export interface RequestConfig {
+  client: Client;
+  method: string;
+  reqOpts?: object;
```
JustinBeckwith: Aren't these `DecorateRequestOptions`?

src/index.ts
```diff
@@ -39,6 +40,81 @@ const {grpc} = new gax.GrpcClient(opts);
...
+ * @internal
+ */
+interface GetClientCallback {
+  /**
+   * @param err - Error, if any.
+   * @param gaxClient - The gax client specified in RequestConfig.client.
+   *                    Typed any since it's importing Javascript source.
+   */
+  // tslint:disable-next-line:no-any
+  (err: Error|null, gaxClient?: any): void;
```
JustinBeckwith: There should be a good definition for the gax client, I just can't remember which. Gax is pretty well typed.

src/index.ts
kinwa91: gRPC library - no `nodejs-common` :(

src/index.ts
kinwa91: Nope - this is a rather custom object unique to the PubSub library as a hybrid (gapic + ts-hand-written layer).

`client` is the GAPIC client name (as a`string`), e.g. PublisherClient, exposed in `src/v1/publisher_client.js`
`method` is the method name (as a `string`) of the client to call - e.g. `method: listTopics` would invoke `PublisherClient.listTopics`
`reqOpts` are the options to the rpc method, [example](https://cloud.google.com/pubsub/docs/reference/rpc/google.pubsub.v1#google.pubsub.v1.GetTopicRequest), which depends on which method it is.
`gaxOpts` are configuration to gax and gRPC (autoPaginate, retry etc..).

src/index.ts
kinwa91: [`gax.ClientStub`](https://github.com/googleapis/gax-nodejs/blob/master/src/grpc.ts#L107)

PR #398, opened 2018-12-07
Add migration for v0.20.0 from v0.19.0

crwilcox:
It was reported that when upgrading from v0.19 to v0.20 an error about requiring new was exposed. Adding a note about this change to the changelog

- [ ] Tests and linter pass


JustinBeckwith:
Thanks :). Can you add this change to the GitHub release as well?

PR #399, opened 2018-12-07
fix(ts): copy gapic code properly

jkwlui:


package.json
```diff
@@ -42,7 +42,7 @@
     "fix": "eslint --fix '**/*.js' && gts fix",
     "generate-scaffolding": "repo-tools generate all && repo-tools generate lib_samples_readme -l samples/ --config ../.cloud-repo-tools.json",
     "clean": "gts clean",
-    "compile": "tsc -p . && cp -r src/v1/ build/src/v1/ && cp -r protos build/ && cp smoke-test/*.js build/system-test && cp test/*.js build/test",
+    "compile": "tsc -p . && cp -r src/v1 build/src/ && cp -r protos build/ && cp smoke-test/*.js build/system-test && cp test/*.js build/test",
```
kinwa91: When `compile` is run multiple times, the original will copy `v1` inside `build/src/v1`, which results in the duplicate docs we see here: https://cloud.google.com/nodejs/docs/reference/pubsub/0.22.x/v1.SubscriberClient

JustinBeckwith: lol oops

JustinBeckwith (approved):

PR #419, opened 2019-01-10
docs(samples): correct publish retry settings

anguillanneuf:
- corrected retryCodes to a list of Google API canonical error codes
- corrected backOffSettings to default values specified in `src/v1/publisher_client_config.json`

ref: https://googleapis.github.io/gax-nodejs/gax.js.html


kinwa91:
@stephenplusplus @callmehiphop need a closer look into this one. Had a look at gax-nodejs but I wasn’t familiar enough to determine the retry parameter is supplied the way it should be.

callmehiphop:
@kinwa91 If we're just supplying the default retry options for publishing, then these definitely match what is listed in the `publisher_client_config.json`. I'm concerned that these could become out of sync in the future if the config json is regenerated with tweaked values.

anguillanneuf:
@kinwa91 The sample should be ready to be merged now. Thanks!

PR #426, opened 2019-01-15
refactor(topic): localize publisher to topic instance

### LINKED ISSUE #245: API response objects shouldn't use arrays to represent structured data
https://cloud.google.com/nodejs/docs/reference/pubsub/0.19.x/global

Many response types are arrays rather than objects. For example: CreateTopicResponse is an array of size 2 where the first element is a Topic and the second element is "The full API response"

These should be members of an object rather than elements of array.

### LINKED ISSUE #246: Publisher class is incomplete
https://cloud.google.com/nodejs/docs/reference/pubsub/0.19.x/Publisher
Publisher has 1 method: publish (which supports batching)

https://cloud.google.com/nodejs/docs/reference/pubsub/0.19.x/v1.PublisherClient
v1.Publisher client has over a dozen methods, including publish (which does not support batching)

Can these APIs be combined? It would be ideal if v1.PublisherClient's publish method just supported batching, and the Publisher class could just be removed from the API

callmehiphop:
Fixes #246

- [x] Tests and linter pass

---

### Breaking change ahead!

We're already about to make a semver major release, so I figured we might as well sneak this in if we can.

#### Before

```js
const publisher = topic.publisher();
await publisher.publish(Buffer.from('Hello, world!'));
```

#### After

```js
await topic.publish(Buffer.from('Hello, world!'));
```

### DIFF
--- a/samples/system-test/subscriptions.test.js
+++ b/samples/system-test/subscriptions.test.js
@@ -109,17 +109,13 @@ describe('subscriptions', () => {
   it('should listen for messages', async () => {
     const messageIds = await pubsub
       .topic(topicNameOne)
-      .publisher()
       .publish(Buffer.from(`Hello, world!`));
     const output = await exec(`${cmd} listen-messages ${subscriptionNameOne}`);
     assert.match(output, new RegExp(`Received message ${messageIds}:`));
   });
 
   it('should listen for messages synchronously', async () => {
-    pubsub
-      .topic(topicNameOne)
-      .publisher()
-      .publish(Buffer.from(`Hello, world!`));
+    pubsub.topic(topicNameOne).publish(Buffer.from(`Hello, world!`));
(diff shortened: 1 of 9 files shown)


callmehiphop:
@jganetsk I can't add you as a reviewer, but PTAL!

JustinBeckwith (approved):
LGTM assuming tests pass :)

jganetsk:
I assume the Publisher class I see in the diff is an implementation detail, and not part of the public API?

callmehiphop:
@jganetsk correct! Users won't interact with it directly, but instead they'll call `topic.publish()` and if they want to configure it later, they can call `topic.setOptions`, which in hindsight maybe we should call `setPublisherOptions` to avoid any confusion.

jganetsk:
LGTM

PR #428, opened 2019-01-16
Return the promises from subscriber instance

Calvin-Huang:
- [X] Tests and linter pass
- [X] Code coverage does not decrease (if any source code was changed)

For example, the `ack` method of `Message` class return the subscriber's ack when it is not handled. We should response a promise just like what we get from subscriber's ack.

### DIFF
--- a/samples/subscriptions.js
+++ b/samples/subscriptions.js
@@ -244,14 +244,14 @@ function listenForMessages(subscriptionName, timeout) {
 
   // Create an event handler to handle messages
   let messageCount = 0;
-  const messageHandler = message => {
+  const messageHandler = async message => {
     console.log(`Received message ${message.id}:`);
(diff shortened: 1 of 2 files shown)


callmehiphop:
hi @Calvin-Huang and thanks for the contribution! I definitely see the value in having promise/callback support on Message methods like ack and nack. Unfortunately this isn't the first time this request has been made, but we've had some push back from the PubSub API team in the past. I think our best course of action would be to close this for now and open an issue to try and determine our best course of action is.

WDYT?

Calvin-Huang:
@callmehiphop 
Sure, good to hear that new from you. Lemme check the issues again to ensure the same issue is not raised.
Look forward to any update for enhancing the `ack` behavior and reliability. Thanks!

PR #447, opened 2019-01-25
fix(auth): pass project id to gax clients

### LINKED ISSUE #288: running pm2 spikes cpu and memory
using with
```
pm2: 3.2.2
node: 8.12.0
npm : 6.4.1
@google-cloud/pubsub: 0.20.1
os: Ubuntu 16.04.3 LTS
subscription type: pull
```

file: **subscription.js**
```js
const path = require('path');
const logger = require('debug')(`job:${path.basename(__dirname)}: `);
const PubSub = require('@google-cloud/pubsub');
const messageHandler = require('./handler');
const settings = require('../settings');

const keyFilePath = path.resolve(__dirname, './key-file.json');

// Creates a client
const pubsub = new PubSub({
  keyFilename: keyFilePath,
});

const { subscriptionName } = settings.google;

// References an existing subscription
const subscription = pubsub.subscription(subscriptionName);

logger('listening to any publication to any Businesses Notification');

subscription.on('message', messageHandler);
```

running this in pm2 initially

first
start-1

second
second

third
third

fourth
fourth

is there other debugging environment i should use to deepen the debugging?

thanks.

### LINKED ISSUE #445: Load cpu 100% macOS
**Environment details**
 - OS: MacOs Mojave 
 - Node.js version: 8.12.0 and 10
 - npm version: 10.5.0 and 6.1.0
 - `@google-cloud/pubsub` version: 0.22.0 and 0.23.0

I am having a problem with a pubsub sdk.
I've tried to reproduced it on other pc and have got the same result.
If I try to run a subscribe script, I will get cpu loading on 100%
If i use linux, it will work ok.
```
const { PubSub } = require('@google-cloud/pubsub');
const pubsub = new PubSub({
  credentials: {}
});
const subscription = pubsub.subscription(topicName);
subscription.on(`message`, message => {
  console.log(`Received message ${message.id}:`);
  console.log(`\tData: ${message.data}`);
  console.log(`\tAttributes: ${message.attributes}`);
  messageCount += 1;
  message.ack();
});
```

Screen load cpu

callmehiphop:
Fixes #288
Fixes #445
 
- [x] Tests and linter pass

---

Looks like not passing the project ID into the gapic clients causes a bunch of future `getProjectId` calls which makes CPU usage spike pretty high

```sh
 [Summary]:
   ticks  total  nonlib   name
   3380   67.2%   71.3%  JavaScript

 [Bottom up (heavy) profile]:
   ticks parent  name
    922   18.3%  LazyCompile: *auth.getProjectId /Users/taco/projects/temp/node_modules/@google-cloud/pubsub/build/src/index.js:599:36
    921   99.9%    LazyCompile: *getProjectIdAsync.then.r /Users/taco/projects/temp/node_modules/google-auth-library/build/src/auth/googleauth.js:73:43
    921  100.0%      T node::RunMicrotasks(v8::FunctionCallbackInfo<v8::Value> const&)
    921  100.0%        LazyCompile: ~_tickCallback internal/process/next_tick.js:41:25
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -806,6 +806,8 @@ export class PubSub {
           return;
         }
         this.projectId = projectId!;
+        this.options.projectId = projectId!;
+
         this.getClient_(config, callback);
       });
       return;
--- a/test/index.ts
+++ b/test/index.ts
@@ -1084,6 +1084,7 @@ describe('PubSub', () => {
         pubsub.getClient_(CONFIG, err => {
           assert.ifError(err);
           assert.strictEqual(pubsub.projectId, PROJECT_ID);
+          assert.strictEqual(pubsub.options.projectId, PROJECT_ID);
           done();
         });
       });


stephenplusplus:
Do the GAPIC clients just need to cache the project ID that they fetch? In any case, this PR is correct-- we should provide to the GAPIC what the user provides, which could be different from what the environment detects. Just curious if there's a simple optimization that would make GAPIC perform better.

callmehiphop:
@stephenplusplus The auth client should cache the project ID (at least it looks like it does). When a gapic client is created it makes an [`auth.getClient`](https://github.com/googleapis/google-auth-library-nodejs/blob/master/src/auth/googleauth.ts#L675) call for every [service stub](https://github.com/googleapis/nodejs-pubsub/blob/master/src/v1/subscriber_client.js#L158) it creates. I haven't looked to deep into this, but it seems like we might be hitting the auth client with concurrent requests for credentials, which might explain the spike in CPU usage. If I had to guess, I think [caching the auth client](https://github.com/googleapis/gax-nodejs/blob/master/src/grpc.ts#L170) would help. Alternatively I think we could also pass our auth client to the gapic client, at which point the credentials would already be cached.

Issue #313, opened 2018-11-02
Enable noImplicitAny in the tsconfig


JustinBeckwith:
The goal here is to provide perfect TypeScript types for customers using the module.  Generally when doing this work I like to:

1. enable `noImplicitAny` in the tsconfig
2. Choose a single file to get passing, and fix all errors in that file (using VSCode here is really useful)
3. disable`noImplicitAny`
4. Get all the tests passing
5. Submit a PR for just the single file I updated
6. Repeat until finished

It's *a lot* easier if we do a PR per file on this one.

praveenqlogic:
#https://github.com/googleapis/nodejs-pubsub/pull/457

PR #457, opened 2019-01-31
refactor(typescript): noImplicityAny for snapshot.ts and publisher.ts

praveenqlogic:

steps
enable noImplicityAny, for snapshot.ts and publisher.ts file in SRC folder!

- [ ] Tests and linter pass

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -45,6 +45,8 @@ import {BatchPublishOptions} from './publisher';
 const opts = {} as gax.GrpcClientOptions;
 const {grpc} = new gax.GrpcClient(opts);
 
+export type SeekCallback = RequestCallback<google.pubsub.v1.ISeekResponse>;
+
 export interface GetSubscriptionMetadataCallback {
   (err: ServiceError|null, res?: google.pubsub.v1.Subscription|null): void;
 }
--- a/src/index.ts
+++ b/src/index.ts
@@ -48,7 +48,7 @@ const {grpc} = new gax.GrpcClient(opts);
 export type SeekCallback = RequestCallback<google.pubsub.v1.ISeekResponse>;
 
 export interface GetSubscriptionMetadataCallback {
-  (err: ServiceError|null, res?: google.pubsub.v1.Subscription|null): void;
+  (err: ServiceError|null, res?: google.pubsub.v1.ISubscription|null): void;
 }
 
 export interface ExistsCallback {
@@ -84,7 +84,7 @@ export interface SubscriptionCallOptions {
  */
 export interface CreateSnapshotCallback {
   (err: Error|null, snapshot?: Snapshot|null,
-   apiResponse?: google.pubsub.v1.Snapshot): void;
+   apiResponse?: google.pubsub.v1.ISnapshot): void;
 }
 
 /**
@@ -92,7 +92,7 @@ export interface CreateSnapshotCallback {
  * @property {Snapshot}.
  * @property {object} 1 The full API response.
  */
-export type CreateSnapshotResponse = [Snapshot, google.pubsub.v1.Snapshot];
+export type CreateSnapshotResponse = [Snapshot, google.pubsub.v1.ISnapshot];
 
 /**
  * Project ID placeholder.
@@ -109,7 +109,7 @@ export type Metadata = any;
  * @property {Topic} 0 The new {@link Topic}.
  * @property {object} 1 The full API response.
  */
-export type CreateTopicResponse = [Topic, google.pubsub.v1.Topic];
+export type CreateTopicResponse = [Topic, google.pubsub.v1.ITopic];
 
 /**
  * @callback CreateTopicCallback
@@ -119,7 +119,7 @@ export type CreateTopicResponse = [Topic, google.pubsub.v1.Topic];
  */
 export interface CreateTopicCallback {
   (err?: Error|null, topic?: Topic|null,
-   apiResponse?: google.pubsub.v1.Topic): void;
+   apiResponse?: google.pubsub.v1.ITopic): void;
 }
 
 /**
@@ -130,7 +130,7 @@ export interface CreateTopicCallback {
  */
 export interface CreateSubscriptionCallback {
   (err?: Error|null, subscription?: Subscription|null,
-   apiResponse?: google.pubsub.v1.Subscription): void;
+   apiResponse?: google.pubsub.v1.ISubscription): void;
 }
 
 export type Client = 'PublisherClient'|'SubscriberClient';
@@ -157,7 +157,7 @@ export interface RequestCallback<TResponse> {
  * @property {object} 1 The full API response.
  */
 export type CreateSubscriptionResponse =
-    [Subscription, google.pubsub.v1.Subscription];
+    [Subscription, google.pubsub.v1.ISubscription];
 
 
 export interface CreateSubscriptionOptions {
@@ -403,7 +403,7 @@ export class PubSub {
       name: subscription.name,
     });
 
-    this.request<google.pubsub.v1.Subscription>(
+    this.request<google.pubsub.v1.ISubscription>(
         {
           client: 'SubscriberClient',
           method: 'createSubscription',
--- a/src/index.ts
+++ b/src/index.ts
@@ -40,15 +40,10 @@ import {Readable} from 'stream';
 import {google} from '../proto/pubsub';
 import {ServiceError} from 'grpc';
 import {FlowControlOptions} from './lease-manager';
+import {BatchPublishOptions} from './publisher';
 
 const opts = {} as gax.GrpcClientOptions;
 const {grpc} = new gax.GrpcClient(opts);
-
-
-export interface SnapshotParent extends PubSub {
-  createSnapshot?: Function;
-  seek?: Function;
-}
 
 export interface GetSubscriptionMetadataCallback {
   (err: ServiceError|null, res?: google.pubsub.v1.Subscription|null): void;
@@ -60,24 +55,6 @@ export interface ExistsCallback {
 
 export interface GetCallOptions extends CallOptions {
   autoCreate?: boolean;
-}
-
-export interface Inventory {
-  callbacks?: Array<RequestCallback<string>>;
-  queued?: Array<{}>;
-  bytes: number;
-  nack?: Array<[string, number]>;
-}
-
-export interface Batching {
-  maxBytes?: number;
-  maxMessages: number;
-  maxMilliseconds?: number;
-}
-
-export interface PublisherCallOptions {
-  batching: Batching;
-  gaxOpts?: CallOptions;
 }
 
 export interface Attributes {
@@ -92,7 +69,7 @@ export interface SubscriptionCallOptions {
   ackDeadline?: number;
   autoPaginate?: boolean;
   gaxOpts?: CallOptions;
-  batching?: Batching;
+  batching?: BatchPublishOptions;
 }
 
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -33,16 +33,22 @@ const PKG = require('../../package.json');
 const v1 = require('./v1');
 
 import {Snapshot} from './snapshot';
-import {Subscription, SubscriptionMetadata, SubscriptionMetadataRaw} from './subscription';
+import {Subscription, SubscriptionMetadataRaw} from './subscription';
 import {Topic, PublishOptions} from './topic';
 import {CallOptions} from 'google-gax';
 import {Readable} from 'stream';
 import {google} from '../proto/pubsub';
 import {ServiceError} from 'grpc';
+import {FlowControlOptions} from './lease-manager';
 
 const opts = {} as gax.GrpcClientOptions;
 const {grpc} = new gax.GrpcClient(opts);
 
+
+export interface SnapshotParent extends PubSub {
+  createSnapshot?: Function;
+  seek?: Function;
+}
 
 export interface GetSubscriptionMetadataCallback {
   (err: ServiceError|null, res?: google.pubsub.v1.Subscription|null): void;
@@ -56,28 +62,40 @@ export interface GetCallOptions extends CallOptions {
   autoCreate?: boolean;
 }
 
-export interface PushConfig {
-  pushEndpoint: string;
-  attibutes?: Map<string, string>;
+export interface Inventory {
+  callbacks?: Array<RequestCallback<string>>;
+  queued?: Array<{}>;
+  bytes: number;
+  nack?: Array<[string, number]>;
+}
+
+export interface Batching {
+  maxBytes?: number;
+  maxMessages: number;
+  maxMilliseconds?: number;
+}
+
+export interface PublisherCallOptions {
+  batching: Batching;
+  gaxOpts?: CallOptions;
+}
+
+export interface Attributes {
+  [key: string]: string;
 }
 
 
 export interface SubscriptionCallOptions {
-  flowControl?:
-      {maxBytes?: number, maxMessages?: number, allowExcessMessages: boolean;};
+  flowControl?: FlowControlOptions;
   maxConnections?: number;
   topic?: Topic;
   ackDeadline?: number;
   autoPaginate?: boolean;
   gaxOpts?: CallOptions;
-  batching?:
-      {maxBytes?: number, maxMessages?: number, maxMilliseconds?: number};
+  batching?: Batching;
 }
 
-export interface PublisherCallOptions {
-  batching?:
-      {maxBytes?: number, maxMessages?: number, maxMilliseconds?: number};
-}
+
 
 /**
  * @callback CreateTopicCallback
@@ -164,7 +182,7 @@ export type CreateSubscriptionResponse =
 
 
 export interface CreateSubscriptionOptions {
-  flowControl?: {maxBytes?: number; maxMessages?: number;};
+  flowControl?: FlowControlOptions;
   gaxOpts?: CallOptions;
   /**
    * Duration in seconds.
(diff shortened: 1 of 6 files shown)


JustinBeckwith:
Greetings!  Drop me a note when the tests are passing and I'll take a look :)

test/snapshot.ts
```diff
@@ -84,8 +86,9 @@ describe('Snapshot', () => {
       assert(promisified);
     });
 
-    it('should localize parent.Promise', () => {
-      assert.strictEqual(snapshot.Promise, SUBSCRIPTION.Promise);
+    it('should localize parent.Promise', () => {  
+      snapshot = new Snapshot(PUBSUB,SNAPSHOT_NAME);
```
AVaksman: to pass this test, please use a real `PubSub` object, 
```javascript
const pubsub = new PubSub();
snapshot = new Snapshot(pubsub, SNAPSHOT_NAME)
```

test/snapshot.ts
```diff
@@ -40,19 +42,19 @@ describe('Snapshot', () => {
   const PROJECT_ID = 'grape-spaceship-123';
 
   const PUBSUB = {
+    Promise:{},
```
AVaksman: This is not needed anymore as the real PubSub class should be used in the test below.

src/index.ts
```diff
@@ -33,17 +33,23 @@ const PKG = require('../../package.json');
...
 import {Readable} from 'stream';
 import {google} from '../proto/pubsub';
 import {ServiceError} from 'grpc';
+import {FlowControlOptions} from './lease-manager';
 
 const opts = {} as gax.GrpcClientOptions;
 const {grpc} = new gax.GrpcClient(opts);
 
 
+export interface SnapshotParent extends PubSub {
```
callmehiphop: I don't think we should export something like this, a Snapshot's parent can either be a Subscription or the PubSub instance, we don't have any kind of class that mixes the two.

src/index.ts
```diff
@@ -56,28 +62,40 @@ export interface GetCallOptions extends CallOptions {
   autoCreate?: boolean;
 }
 
-export interface PushConfig {
-  pushEndpoint: string;
-  attibutes?: Map<string, string>;
+export interface Inventory {
```
callmehiphop: Pretty sure this is a private member of the Publisher, so I don't think we should be exporting this and it should probably live in the Publisher class.

src/index.ts
```diff
@@ -56,28 +62,40 @@ export interface GetCallOptions extends CallOptions {
...
+  callbacks?: Array<RequestCallback<string>>;
+  queued?: Array<{}>;
+  bytes: number;
+  nack?: Array<[string, number]>;
+}
+
+export interface Batching {
```
callmehiphop: This interface already exists https://github.com/googleapis/nodejs-pubsub/blob/master/src/publisher.ts#L44

src/index.ts
```diff
@@ -56,28 +62,40 @@ export interface GetCallOptions extends CallOptions {
...
+  maxBytes?: number;
+  maxMessages: number;
+  maxMilliseconds?: number;
+}
+
+export interface PublisherCallOptions {
```
callmehiphop: This interface already exists https://github.com/googleapis/nodejs-pubsub/blob/master/src/publisher.ts#L56

src/publisher.ts
```diff
@@ -157,17 +154,21 @@ export class Publisher {
...
+      data: Buffer, attributesOrCallback?: Attributes|PublishCallback,
+      callback?: PublishCallback): Promise<string>|void {
     if (!(data instanceof Buffer)) {
       throw new TypeError('Data must be in the form of a Buffer.');
     }
-    if (is.fn(attributes)) {
-      callback = attributes;
-      attributes = {};
-    }
+    const attributes =
```
callmehiphop: What version of `@google-cloud/common` is tied to this? Can we use [maybeOptionsOrCallback](https://github.com/googleapis/nodejs-common/blob/master/src/util.ts#L756)?

src/index.ts
```diff
@@ -56,28 +62,40 @@ export interface GetCallOptions extends CallOptions {
   autoCreate?: boolean;
 }
 
-export interface PushConfig {
-  pushEndpoint: string;
-  attibutes?: Map<string, string>;
+export interface Inventory {
+  callbacks?: Array<RequestCallback<string>>;
+  queued?: Array<{}>;
```
callmehiphop: `callbacks` and `queued` should not be optional, they will always be set.

src/index.ts
```diff
@@ -56,28 +62,40 @@ export interface GetCallOptions extends CallOptions {
...
+  bytes: number;
+  nack?: Array<[string, number]>;
```
callmehiphop: I think this found its way here by mistake, there isn't a `nack` array on the publisher inventory.

src/publisher.ts
```diff
@@ -17,19 +17,16 @@
 import {promisifyAll} from '@google-cloud/promisify';
 import * as arrify from 'arrify';
 import {CallOptions} from 'google-gax';
-import {ServiceError} from 'grpc';
```
callmehiphop: Why are we removing this? Seems like if an error occurred we would want the user to be able to inspect the status code/metadata.

src/publisher.ts
```diff
@@ -17,19 +17,16 @@
...
 import {Topic} from './topic';
+import {Inventory, RequestCallback, Attributes} from '.';
 
 export interface PublishCallback {
-  (err: null|ServiceError, messageId: string): void;
-}
-
-interface PublishApiResponse {
-  messageIds: string[];
+  (err?: null|Error, messageId?: string|null): void;
```
callmehiphop: Can we change this to

```ts
export interface PublishCallback {
  (err: ServiceError, messageId?: null): void;
  (err: null, messageId: string): void;
}
```

src/publisher.ts
```diff
@@ -185,10 +186,10 @@ export class Publisher {
       this.publish_();
     }
     // add it to the queue!
-    this.queue_(data, attributes, callback);
+    this.queue_(data, attributes, callback!);
     // next lets check if this message brings us to the message cap or if we
     // hit the max byte limit
-    const hasMaxMessages = this.inventory_.queued.length === opts.maxMessages!;
+    const hasMaxMessages = this.inventory_.queued!.length === opts.maxMessages;
```
callmehiphop: Shouldn't need the`!` here.

src/publisher.ts
```diff
@@ -256,11 +257,14 @@ export class Publisher {
...
-          each(callbacks, (callback, next) => {
-            const messageId = messageIds[callbacks.indexOf(callback)];
-            callback(err, messageId);
-            next();
-          });
+          each(
+              callbacks,
+              (callback: RequestCallback<string>,
+               next: RequestCallback<string>) => {
+                const messageId = messageIds[callbacks!.indexOf(callback)];
```
callmehiphop: Shouldn't need the `!` here.

src/publisher.ts
```diff
@@ -272,13 +276,17 @@ export class Publisher {
...
    * @param {function} callback The callback function.
    */
-  queue_(data, attrs, callback) {
-    this.inventory_.queued.push({
+  queue_(data: Buffer, attrs: Attributes): Promise<string>;
+  queue_(data: Buffer, attrs: Attributes, callback: RequestCallback<string>):
+      void;
+  queue_(data: Buffer, attrs: Attributes, callback?: RequestCallback<string>):
+      void|Promise<string> {
+    this.inventory_.queued!.push({
```
callmehiphop: Shouldn't need the `!` here.

src/publisher.ts
```diff
@@ -272,13 +276,17 @@ export class Publisher {
...
       data,
       attributes: attrs,
     });
     this.inventory_.bytes += data.length;
-    this.inventory_.callbacks.push(callback);
+    this.inventory_.callbacks!.push(callback!);
```
callmehiphop: Shouldn't need the `!` here.

src/snapshot.ts
```diff
@@ -207,8 +144,42 @@ export class Snapshot {
...
+  seek(
+      gaxOpts: CallOptions,
+      callback: google.pubsub.v1.Subscriber.SeekCallback): void;
+  seek(
+      gaxOpts?: CallOptions|google.pubsub.v1.Subscriber.SeekCallback,
+      callback?: google.pubsub.v1.Subscriber.SeekCallback):
+      void|Promise<google.pubsub.v1.SeekResponse> {
+    if (!(this.parent instanceof Subscription)) {
+      throw new Error(
+          `This is only available if you accessed this object through {@link Subscription#snapshot}`);
```
callmehiphop: Lets drop the JSDoc portion of the error message and just return `Subscription#snapshot`

test/snapshot.ts
```diff
@@ -117,23 +120,29 @@ describe('Snapshot', () => {
...
-        };
+        sandbox.stub(subscription, 'createSnapshot')
+            .callsFake((name: string) => {
+              assert.strictEqual(name, FULL_SNAPSHOT_NAME);
+              done();
+            });
 
-        const snapshot = new Snapshot(SUBSCRIPTION, SNAPSHOT_NAME);
+        const snapshot = new Snapshot(subscription, SNAPSHOT_NAME);
         snapshot.create(done);
```
callmehiphop: Should probably pass `assert.ifError` instead of `done` since we're ignoring the callback inside of the stub.

test/snapshot.ts
```diff
@@ -117,23 +120,29 @@ describe('Snapshot', () => {
...
-        };
-
-        const snapshot = new Snapshot(SUBSCRIPTION, SNAPSHOT_NAME);
+      it('should call the seek method', done => {
+        sandbox.stub(subscription, 'seek').callsFake((snapshot) => {
+          assert.strictEqual(snapshot, FULL_SNAPSHOT_NAME);
+          done();
+        });
+        const snapshot = new Snapshot(subscription, SNAPSHOT_NAME);
         snapshot.seek(done);
```
callmehiphop: Should probably pass assert.ifError instead of done since we're ignoring the callback inside of the stub.

test/snapshot.ts
praveenqlogic: done

src/index.ts
praveenqlogic: Removed, i missed while merging, Thanks 👍

src/index.ts
praveenqlogic: done

src/publisher.ts
praveenqlogic: nodejs-pubsub is not using @google-cloud/common.

src/index.ts
praveenqlogic: done

src/index.ts
praveenqlogic: removed.

praveenqlogic:
@JustinBeckwith Hi, tests are passing, please take a look into it.

src/publisher.ts
```diff
@@ -17,19 +17,22 @@
...
 import * as extend from 'extend';
 import * as is from 'is';
 import {Topic} from './topic';
+import {RequestCallback, Attributes} from '.';
 
-export interface PublishCallback {
-  (err: null|ServiceError, messageId: string): void;
+interface Inventory {
+  callbacks: Array<RequestCallback<string>>;
+  queued: Array<{}>;
```
callmehiphop: I think we could probably change this to whatever the equivalent of [PubsubMessage](https://github.com/googleapis/nodejs-pubsub/blob/master/protos/google/pubsub/v1/pubsub.proto#L384) is.

src/snapshot.ts
```diff
@@ -185,18 +118,22 @@ export class Snapshot {
...
     };
     callback = callback || util.noop;
-    this.parent.request(
-        {
-          client: 'SubscriberClient',
-          method: 'deleteSnapshot',
-          reqOpts,
-        },
-        callback);
+    (this.parent as PubSub)
```
callmehiphop: Can we do something to remove the cast here? Seems like if we just improved the type of whatever `Subscription#request` is that this wouldn't be needed.

src/snapshot.ts
```diff
@@ -207,8 +144,40 @@ export class Snapshot {
...
+
+
+  create(gaxOpts?: CallOptions): Promise<CreateSnapshotResponse>;
+  create(callback: CreateSnapshotCallback): void;
+  create(gaxOpts: CallOptions, callback: CreateSnapshotCallback): void;
+  create(
+      gaxOpts?: CallOptions|CreateSnapshotCallback,
+      callback?: CreateSnapshotCallback): void|Promise<CreateSnapshotResponse> {
+    if (!(this.parent instanceof Subscription)) {
+      throw new Error(`Subscription#snapshot`);
```
callmehiphop: Sorry for any miscommunication on my part, but what I meant when I said drop the JSDoc part was just removing the `{@link}` tag. We should keep the beginning part that said `This is only available if you accessed this object through`.

src/snapshot.ts
```diff
@@ -207,8 +144,40 @@ export class Snapshot {
...
+  seek(callback: google.pubsub.v1.Subscriber.SeekCallback): void;
+  seek(
+      gaxOpts: CallOptions,
+      callback: google.pubsub.v1.Subscriber.SeekCallback): void;
+  seek(
+      gaxOpts?: CallOptions|google.pubsub.v1.Subscriber.SeekCallback,
+      callback?: google.pubsub.v1.Subscriber.SeekCallback):
+      void|Promise<google.pubsub.v1.SeekResponse> {
+    if (!(this.parent instanceof Subscription)) {
+      throw new Error(`Subscription#snapshot`);
```
callmehiphop: See my above comment about throwing for `Snapshot#create`

src/snapshot.ts
```diff
@@ -207,8 +144,40 @@ export class Snapshot {
...
+      callback?: CreateSnapshotCallback): void|Promise<CreateSnapshotResponse> {
+    if (!(this.parent instanceof Subscription)) {
+      throw new Error(`Subscription#snapshot`);
+    }
+    return (this.parent as Subscription)
+        .createSnapshot(this.name, gaxOpts! as CallOptions, callback!);
+  }
+
+  seek(gaxOpts?: CallOptions): Promise<google.pubsub.v1.SeekResponse>;
+  seek(callback: google.pubsub.v1.Subscriber.SeekCallback): void;
```
callmehiphop: I'm not sure if we should be using the generated callbacks (anywhere ever). They all look to return the proto class version of the response which we don't use. I think instead we'll want something like

```ts
type SeekCallback = RequestCallback<google.pubsub.v1.Subscriber.ISeekResponse>;
```

src/snapshot.ts
```diff
@@ -85,85 +86,17 @@ import * as util from './util';
...
-  constructor(parent, name: string) {
-    if (parent.Promise) {
+  metadata!: google.pubsub.v1.Snapshot;
+  constructor(parent: Subscription|PubSub, name: string) {
+    if (parent instanceof PubSub) {
       this.Promise = parent.Promise;
     }
     this.parent = parent;
     this.name = Snapshot.formatName_(parent.projectId, name);
-    if (is.fn(parent.createSnapshot)) {
```
JustinBeckwith: Feels like this requires a bit of an explanation :) What's up here?

JustinBeckwith:
I totally defer to @callmehiphop on this one 👍

src/snapshot.ts
callmehiphop: For whatever reason the PubSub API allows you to list all snapshots for a project without providing information about the parent subscription. This gets kind of funky because you need to know the subscription name in order to create or seek a snapshot. In the event that the user got a snapshot via `PubSub#snapshot` or `PubSub#getSnapshots` then `create` and `seek` simply would not exist. The changes here just make it so that instead you can call those methods, but they will return an error stating why they aren't available.

src/publisher.ts
praveenqlogic: Done!

src/snapshot.ts
praveenqlogic: Done

src/snapshot.ts
praveenqlogic: Done

src/publisher.ts
```diff
@@ -256,11 +271,14 @@ export class Publisher {
...
         (err, resp) => {
           const messageIds = arrify(resp && resp.messageIds);
-          each(callbacks, (callback, next) => {
-            const messageId = messageIds[callbacks.indexOf(callback)];
-            callback(err, messageId);
-            next();
-          });
+          each(
+              callbacks,
+              (callback: RequestCallback<string>,
```
callmehiphop: This can be `PublishCallback`

src/publisher.ts
```diff
@@ -247,7 +262,7 @@ export class Publisher {
       topic: this.topic.name,
       messages,
     };
-    this.topic.request(
+    this.topic.request<google.pubsub.v1.PublishResponse>(
```
callmehiphop: This is kind of a general note about using the generated types. I'm pretty sure we are only ever going to want to use the interface versions of these. The generated callbacks and classes all reference gRPC reflection objects, which gax and grpc do not provide for us by default. There aren't very many differences between the class/interface versions, so this *might* fly under the radar, but if the interface version is available to us then I think we should use that instead.

/cc @JustinBeckwith @stephenplusplus @AVaksman

src/publisher.ts
```diff
@@ -256,11 +266,12 @@ export class Publisher {
         },
         (err, resp) => {
           const messageIds = arrify(resp && resp.messageIds);
-          each(callbacks, (callback, next) => {
-            const messageId = messageIds[callbacks.indexOf(callback)];
-            callback(err, messageId);
-            next();
-          });
+          each(
+              callbacks, (callback: PublishCallback, next: PublishCallback) => {
```
callmehiphop: Oof, missed this, but `next` should probably just be `Function`.

praveenqlogic: done

praveenqlogic:
#313

PR #482, opened 2019-02-15
refactor(typescript): noImplicityAny for index.ts

### LINKED ISSUE #313: Enable noImplicitAny in the tsconfig

praveenqlogic:
Fixes #313 (it's a good idea to open an issue first for discussion)

- [ ] Tests and linter pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
   },
   "repository": "googleapis/nodejs-pubsub",
   "main": "./build/src/index.js",
+  "types": "./build/src/index.d.ts",
   "files": [
     "build/protos",
     "build/src",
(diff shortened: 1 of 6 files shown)


praveenqlogic:
locally both ('should get a list of snapshots','should get a list of snapshots as a stream') test case is passing.

JustinBeckwith:
👋 @callmehiphop looks like another potential case of a flake

PR #487, opened 2019-02-18
refactor(typescript): noImplicityAny for Topic test file

praveenqlogic:
refactor(typescript) changes for #313 (it's a good idea to open an issue first for discussion)

- [ ] Tests and linter pass

### DIFF
--- a/test/topic.ts
+++ b/test/topic.ts
@@ -16,14 +16,18 @@
 
 import * as pfy from '@google-cloud/promisify';
 import * as assert from 'assert';
+import {CallOptions} from 'google-gax';
+import {ServiceError} from 'grpc';
 import * as proxyquire from 'proxyquire';
 import * as sinon from 'sinon';
 
+import {google} from '../proto/pubsub';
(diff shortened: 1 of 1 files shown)


test/topic.ts
```diff
@@ -82,7 +86,7 @@ describe('Topic', () => {
   // tslint:disable-next-line no-any variable-name
   let Topic: any;
   // tslint:disable-next-line no-any
```
JustinBeckwith: You can delete this line now

test/topic.ts
```diff
@@ -164,7 +168,7 @@ describe('Topic', () => {
     });
 
     it('should localize the request function', done => {
-      PUBSUB.request = callback => {
+      PUBSUB.request = (callback: RequestCallback<google.protobuf.Empty>) => {
```
JustinBeckwith: I mean, that or just `Function` here is fine to be honest.  I'm ok with the types being a tad sloppy in the unit tests.

PR #490, opened 2019-02-19
refactor(typescript): noImplicitAny for snapshot and subscriber test file

praveenqlogic:

- [ ] Tests and linter pass

### DIFF
--- a/test/snapshot.ts
+++ b/test/snapshot.ts
@@ -19,8 +19,8 @@ import * as assert from 'assert';
 import * as proxyquire from 'proxyquire';
 import * as sinon from 'sinon';
 
-import {google} from '../proto/pubsub';
-import {PubSub, RequestCallback, RequestConfig, Subscription} from '../src';
+import {PubSub, RequestConfig, Subscription} from '../src';
+import * as snapTypes from '../src/snapshot';
 import * as util from '../src/util';
 
 let promisified = false;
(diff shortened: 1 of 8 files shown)


test/snapshot.ts
```diff
@@ -19,24 +19,25 @@ import * as assert from 'assert';
...
       promisified = true;
     }
   },
 });
 
 describe('Snapshot', () => {
-  // tslint:disable-next-line variable-name
-  let Snapshot;
-  let snapshot;
+  // tslint:disable-next-line variable-name no-any
```
JustinBeckwith: Instead of `any` here, can you do something like:
```js
import * as snapTypes from '../src/snapshot';
let Snapshot: typeof snapTypes.Snapshot:
let snapshot: Snapshot;
```

Note that `snapshot` shouldn't be `typeof` here

test/subscriber.ts
```diff
@@ -39,17 +41,10 @@ interface ClientCallback {
   (error: null|Error, client: FakeClient): void;
 }
 
-class FakePubSub {
```
JustinBeckwith: behavior changes in type refactoring makes me nervous.  Can you leave this the way it was?

test/subscriber.ts
```diff
@@ -381,7 +377,7 @@ describe('Subscriber', () => {
 
       const [options] = spy.lastCall.args;
       assert.deepStrictEqual(options, {client: 'SubscriberClient'});
-      assert.strictEqual(client, pubsub.client);
+      // assert.strictEqual(client, pubsub);
```
JustinBeckwith: That seems bad :)

callmehiphop:
Sort of a general statement - I think if we switched to using `sinon.stub()` instead of overwriting test functions, you won't need to redeclare the param types.

test/subscriber.ts
callmehiphop: You could probably just do something like

```ts
pubsub = new FakePubSub() as unknown as PubSub;
```

test/subscriber.ts
```diff
@@ -129,7 +124,8 @@ describe('Subscriber', () => {
 
   const fakeProjectify = {replaceProjectIdToken: sandbox.stub()};
 
-  let subscription;
+  // tslint:disable-next-line no-any
```
callmehiphop: Do we need this?

test/subscriber.ts
```diff
@@ -371,13 +377,12 @@ describe('Subscriber', () => {
 
   describe('getClient', () => {
     it('should get a subscriber client', async () => {
-      const pubsub = subscription.pubsub;
+      const pubsub = subscription.pubsub as {} as FakePubSub;
```
callmehiphop: Why do we need to cast this?

praveenqlogic: to get access to client, as its a type of FakePubsub and client property is not available in actual pubsub.

PR #491, opened 2019-02-20
Release @google-cloud/pubsub v0.25.0

callmehiphop:
This pull request was generated using releasetool.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,42 @@
 
 [1]: https://www.npmjs.com/package/nodejs-pubsub?activeTab=versions
 
+## v0.25.0
+
+02-20-2019 10:35 PST
+
+### New Features
+- refactor(ts): improve TypeScript types ([#482](https://github.com/googleapis/nodejs-pubsub/pull/482))
+- refactor(typescript): noImplicityAny for snapshot.ts and publisher.ts ([#457](https://github.com/googleapis/nodejs-pubsub/pull/457))
+
+### Bug Fixes
+- fix: ignore messages that come in after close ([#485](https://github.com/googleapis/nodejs-pubsub/pull/485))
+- fix: throw on invalid credentials and update retry config ([#476](https://github.com/googleapis/nodejs-pubsub/pull/476))
+
+### Dependencies
+- chore(deps): update dependency mocha to v6 ([#488](https://github.com/googleapis/nodejs-pubsub/pull/488))
+- fix(deps): update dependency @google-cloud/promisify to ^0.4.0 ([#478](https://github.com/googleapis/nodejs-pubsub/pull/478))
+- fix(deps): update dependency yargs to v13 ([#475](https://github.com/googleapis/nodejs-pubsub/pull/475))
+- fix(deps): update dependency duplexify to v4 ([#462](https://github.com/googleapis/nodejs-pubsub/pull/462))
+- fix(deps): update dependency google-gax to ^0.25.0 ([#456](https://github.com/googleapis/nodejs-pubsub/pull/456))
+
+### Documentation
+- docs: update links in contrib guide ([#479](https://github.com/googleapis/nodejs-pubsub/pull/479))
+- docs: update contributing path in README ([#471](https://github.com/googleapis/nodejs-pubsub/pull/471))
+- chore: move CONTRIBUTING.md to root ([#470](https://github.com/googleapis/nodejs-pubsub/pull/470))
+- docs: make mention of message change in changelog ([#469](https://github.com/googleapis/nodejs-pubsub/pull/469))
+- docs: add lint/fix example to contributing guide ([#464](https://github.com/googleapis/nodejs-pubsub/pull/464))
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,42 @@
...
+
+02-20-2019 10:35 PST
+
+### New Features
+- refactor(ts): improve TypeScript types ([#482](https://github.com/googleapis/nodejs-pubsub/pull/482))
+- refactor(typescript): noImplicityAny for snapshot.ts and publisher.ts ([#457](https://github.com/googleapis/nodejs-pubsub/pull/457))
+
+### Bug Fixes
+- fix: ignore messages that come in after close ([#485](https://github.com/googleapis/nodejs-pubsub/pull/485))
+- fix: throw on invalid credentials and update retry config ([#476](https://github.com/googleapis/nodejs-pubsub/pull/476))
```
crwilcox: I am digging into this particular change. it seems odd that so many things changed from non-idempotent to idempotent. If we can hold this for just a bit that would be super.

CHANGELOG.md
```diff
@@ -4,6 +4,42 @@
 
 [1]: https://www.npmjs.com/package/nodejs-pubsub?activeTab=versions
 
+## v0.25.0
```
JustinBeckwith: doesn't this release having breaking changes?

callmehiphop: Maybe this was a mistake, I'm not really sure how to I should be versioning for TypeScript types. I assumed this was a feature, albeit only for TypeScript users, so I went minor.

callmehiphop: Oh wait, I misunderstood, the retries are breaking, you're right

JustinBeckwith: Yes!  Was just saying - we should call that out :)

CHANGELOG.md
callmehiphop: @jadekler do you have input here? I want to say I remember you telling us this change was coming (apologies if I'm misremembering).

CHANGELOG.md
jadekler: +1 to calling it out, but we're not doing major semver bumps elsewhere if that's part of the question. The "breaking"-ness of the change should be more akin to bug/unintended behavior fixing.

As to calling it out, maybe add something like "If you experience usability concerns, please file an issue at <link>."

callmehiphop: I made a change to include a note about it. I didn't go with the usual **BREAKING** headline, I feel like that should be reserved for surface changes. Let me know if you disagree!

JustinBeckwith (approved):

JustinBeckwith:
@callmehiphop should I be worried about the failing system tests....?

callmehiphop:
@JustinBeckwith great question. I'm less concerned about the grpc-js failures, but I am curious about the sample failure. Its the same test failing pretty consistently via Kokoro and I hadn't seen this particular failure pop up in a little while. Based on the assertion of that test, my best guess is that `setTimeout` is firing much earlier than it should be, do we know if there's any funkiness with timing functions on Kokoro?

JustinBeckwith:
Yeah - please get the samples tests passing reliably before we ship this out.

alexander-fenster:
Please ignore grpc-js tests failures, it's likely has nothing to do with Pub/Sub code, so when gax is fixed, those tests will start passing. I'm looking into it in gax.

jkwlui:
> Based on the assertion of that test, my best guess is that `setTimeout` is firing much earlier than it should be, do we know if there's any funkiness with timing functions on Kokoro?

@callmehiphop I'm not aware of any issue with `setTimeout` on Kokoro. We could eliminate some possibilities by trying to run the samples test in a docker container. The image used in Kokoro is based on `node:8`.

callmehiphop:
@jkwlui that's what I've been trying to do, I took it a step further and deployed my docker image to GCE, but I haven't managed to get the same failure there.

callmehiphop:
Update on my testing, I have managed to reproduce this several times now, but not consistently. My logs tell me that my hunch was correct and that calling `setTimeout(func, 1000)` can fire as soon as 800ms.

I think we may want to explore the idea of using a timing library as opposed to straight up `setTimeout`.

JustinBeckwith:
Is the setTimeout causing trouble here code in a test, or is it in library code?  If it's test code that's misbehaving, I'm supportive of tweaking that code to get the test passing.  

On the specific issue with setTimeout - if the issue was it not invoking the callback within the allotted time, I'd say yeah.  But I'd it's firing *early* - that's a huge bug in node, and we should engage @ofrobots and the node team.

callmehiphop:
@JustinBeckwith its [library code](https://github.com/googleapis/nodejs-pubsub/blob/master/src/publisher.ts#L209). We use a `setTimeout` to buffer messages and send them as a batch. 

As far as calling it a bug in Node.. I'm not sure. It is pretty well known that `setTimeout` can be jittery, but 200ms seems very high.

callmehiphop:
I was incorrect about `setTimeout` being jittery, thanks @alexander-fenster for figuring out the issue.

PR #494, opened 2019-02-22
test: fix publishTime issues

alexander-fenster:
This PR fixes sample test `should publish with specific batch settings` (based on the discussion in #491). I was debugging totally different stuff but realized you're talking about something very interesting and could not resist from debugging this test as well. Good news by the way: sample tests work perfectly with `grpc-js`. Just wanted to let you folks know :)

I added all possible logging to the test (#493 shows my test runs) and found some interesting things that I'd like to share with you folks.

1. `setTimeout` is very accurate. Only once in all my tests it fired 1 ms earlier than needed, and I guess it's more like a time precision problem than the actual early run.

2. The test failure is caused by the sad fact that `publishTime` returned by Pub/Sub is always a _whole second_. Since the test expects millisecond precision, it fails quite often and randomly.

The easiest way to fix a test is to allow up to 1 second offset (accounting for the lost millisecond portion).

See some pictures from my test runs. The first line (`ts=`) is the actual timestamp (logged in `publish_`), the line with `publishTime:` shows the time returned in the message. Note that all `publishTime` timestamps end with `000`.

The `[p] diff` line shows the actual measured sleep time of `setTimeout`. You can see it's correct in both pass and fail test runs.

Fails:

image

image

Passes:

image

image

- [x] Tests and linter pass

### DIFF
--- a/samples/system-test/topics.test.js
+++ b/samples/system-test/topics.test.js
@@ -155,7 +155,7 @@ describe('topics', () => {
   });
 
   it('should publish with specific batch settings', async () => {
-    const waitTime = 1000;
+    const waitTime = 5000;
     const [subscription] = await pubsub
       .topic(topicNameOne)
       .subscription(subscriptionNameThree)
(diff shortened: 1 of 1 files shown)


callmehiphop:
@alexander-fenster dang, nice find! I wonder if this is a PubSub bug, while their grpc docs make no mention of precision, the REST docs say it should be accurate to nanoseconds ([link](https://cloud.google.com/pubsub/docs/reference/rest/v1/PubsubMessage)). That or we are mishandling the protobuf Timestamp.

callmehiphop:
Oh man, turns out it was the test all along! Calling `Date.parse` against a `Date` object actually rounds to the closest second. So the test is just handling `publishTime` poorly.

alexander-fenster:
@callmehiphop Oh! Then I guess you need to fix the test once more :)  Thanks for looking into this!

PR #498, opened 2019-02-26
feat(typescript): ship typescript declaration files

### LINKED ISSUE #141: Convert to TypeScript
Currently blocked on https://github.com/googleapis/gapic-generator/issues/2054

callmehiphop:
Fixes #141


package.json
```diff
@@ -13,8 +13,6 @@
   "files": [
     "build/protos",
```
JustinBeckwith: Going to make a guess - we probably need to include `build/proto` here as well.

callmehiphop: Ah right, good call!

PR #501, opened 2019-02-27
refactor(typescript): noImplicitAny for IAM test file

praveenqlogic:

- [ ] Tests and linter pass

### DIFF
--- a/test/iam.ts
+++ b/test/iam.ts
@@ -18,12 +18,14 @@ import * as promisify from '@google-cloud/promisify';
 import * as assert from 'assert';
 import * as proxyquire from 'proxyquire';
 
+import {PubSub, RequestConfig} from '../src';
+import * as iamTypes from '../src/iam';
 import * as util from '../src/util';
 
 let promisified = false;
 const fakePromisify = Object.assign({}, promisify, {
   // tslint:disable-next-line variable-name
-  promisifyAll(Class) {
+  promisifyAll(Class: typeof iamTypes.IAM) {
     if (Class.name === 'IAM') {
       promisified = true;
     }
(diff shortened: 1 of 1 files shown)


test/iam.ts
```diff
@@ -110,20 +113,14 @@ describe('IAM', () => {
   });
 
   describe('setPolicy', () => {
-    const policy = {etag: 'ACAB'};
-
-    it('should throw an error if a policy is not supplied', () => {
```
JustinBeckwith: We should probably not delete this test

test/iam.ts
```diff
@@ -144,20 +141,14 @@ describe('IAM', () => {
   });
 
   describe('testPermissions', () => {
-    it('should throw an error if permissions are missing', () => {
```
JustinBeckwith: again, why deleting a test?

test/iam.ts
praveenqlogic: in setPolicy function, policy is a required parameter with type Policy
https://github.com/googleapis/nodejs-pubsub/blob/master/src/iam.ts#L234
and after declaring explicit type it will give error, as it will not allow you to do what we are trying to do in this test case due to explicit type declaration (we my use iam as any type than this test may look relevant).

test/iam.ts
praveenqlogic: same scenario as setPolicy function.
https://github.com/googleapis/nodejs-pubsub/blob/master/src/iam.ts#L313

test/iam.ts
praveenqlogic: fix the changes as suggested on another PR
https://github.com/googleapis/nodejs-pubsub/pull/502

test/iam.ts
praveenqlogic: fix the changes as suggested on another PR
https://github.com/googleapis/nodejs-pubsub/pull/502

PR #502, opened 2019-02-27
refactor(typescript):noImplicitAny for index test file

praveenqlogic:

- [ ] Tests and linter pass

### DIFF
--- a/src/topic.ts
+++ b/src/topic.ts
@@ -328,14 +328,8 @@ export class Topic {
         callback!(null, true);
         return;
       }
-      let code = 0;
-      if (err.hasOwnProperty('code')) {
-        code =
-            (Object.getOwnPropertyDescriptor(err, 'code') as PropertyDescriptor)
-                .value;
-      }
-      if (code === 5) {
-        callback!(null, false);
+      if (err.code === 5) {
+        callback(null, false);
         return;
       }
       callback!(err);
@@ -397,13 +391,7 @@ export class Topic {
         callback!(null, this, apiResponse!);
         return;
       }
-      let code = 0;
-      if (err.hasOwnProperty('code')) {
-        code =
-            (Object.getOwnPropertyDescriptor(err, 'code') as PropertyDescriptor)
-                .value;
-      }
-      if (code !== 5 || !autoCreate) {
+      if (err.code !== 5 || !autoCreate) {
         callback!(err, null, apiResponse!);
         return;
       }
--- a/src/topic.ts
+++ b/src/topic.ts
@@ -329,7 +329,7 @@ export class Topic {
         return;
       }
       if (err.code === 5) {
-        callback(null, false);
+        callback!(null, false);
         return;
       }
       callback!(err);
--- a/test/index.ts
+++ b/test/index.ts
@@ -19,14 +19,15 @@ import * as promisify from '@google-cloud/promisify';
 import * as arrify from 'arrify';
 import * as assert from 'assert';
 import * as gax from 'google-gax';
-import {CallOptions} from 'grpc';
+import {CallOptions, ServiceError} from 'grpc';
 import * as proxyquire from 'proxyquire';
 import * as sinon from 'sinon';
 
 import {google} from '../proto/pubsub';
-import * as pubsubTypes from '../src';
+import * as pubsubTypes from '../src/pubsub';
 import {Snapshot} from '../src/snapshot';
 import * as subby from '../src/subscription';
+import {Topic} from '../src/topic';
 import * as util from '../src/util';
 
 
@@ -55,7 +56,7 @@ let subscriptionOverride: any;
 
 function Subscription(
     pubsub: pubsubTypes.PubSub, name: string,
-    options: pubsubTypes.SubscriptionCallOptions) {
+    options: subby.SubscriptionOptions) {
   const overrideFn = subscriptionOverride || subscriptionCached;
   return new overrideFn(pubsub, name, options);
 }
@@ -300,7 +301,7 @@ describe('PubSub', () => {
     pubsub = new pubsubTypes.PubSub({});
     const TOPIC = Object.assign(new FakeTopic(), {
       name: 'projects/' + PROJECT_ID + '/topics/' + TOPIC_NAME,
-    }) as {} as pubsubTypes.Topic;
+    }) as {} as Topic;
 
     const SUB_NAME = 'subscription';
     const SUBSCRIPTION = {
@@ -315,7 +316,7 @@ describe('PubSub', () => {
 
     beforeEach(() => {
       (Subscription as {} as typeof subby.Subscription).formatMetadata_ =
-          (metadata: subby.SubscriptionMetadataRaw) => {
+          (metadata: subby.SubscriptionMetadata) => {
             return Object.assign({}, metadata) as subby.SubscriptionMetadata;
           };
     });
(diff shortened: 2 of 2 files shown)


test/index.ts
```diff
@@ -120,8 +130,9 @@ const v1Override = {};
 // tslint:disable-next-line no-any
 let v1ClientOverrides: any = {};
 
-function defineOverridableClient(clientName) {
+function defineOverridableClient(clientName: string) {
   function DefaultClient() {}
+  /** @type {never[]} */
```
JustinBeckwith: What is that!?

test/index.ts
```diff
@@ -137,13 +148,14 @@ defineOverridableClient('SubscriberClient');
 
 describe('PubSub', () => {
   // tslint:disable-next-line variable-name
-  let PubSub;
+  let PubSub: typeof pubsubTypes.PubSub;
   const PROJECT_ID = 'test-project';
-  let pubsub;
+  // tslint:disable-next-line no-any
```
JustinBeckwith: don't think you need this line

test/index.ts
```diff
@@ -295,60 +309,43 @@ describe('PubSub', () => {
...
     };
 
-    beforeEach(() => {
-      // tslint:disable-next-line no-any
-      (Subscription as any).formatMetadata_ = metadata => {
-        return Object.assign({}, metadata);
-      };
-    });
 
-    it('should throw if no Topic is provided', () => {
```
JustinBeckwith: looks like these two tests got deleted

test/index.ts
```diff
@@ -295,60 +309,43 @@ describe('PubSub', () => {
...
-      };
+      pubsub.request =
+          (config: pubsubTypes.RequestConfig, callback: Function) => {
+            callback(null, apiResponse);
+          };
 
       pubsub.createSubscription(TOPIC, SUB_NAME, done);
     });
 
-    it('should allow undefined/optional configuration options', done => {
```
JustinBeckwith: Why are we deleting a test?

test/index.ts
```diff
@@ -43,17 +49,21 @@ const fakeGoogleGax = {
...
   const overrideFn = subscriptionOverride || subscriptionCached;
-  return new overrideFn(a, b, c);
+  return new overrideFn(pubsub, name, options);
 }
 
 let promisified = false;
 const fakePromisify = Object.assign({}, promisify, {
   // tslint:disable-next-line variable-name
-  promisifyAll(Class, options) {
+  promisifyAll(Class: Function, options: promisify.PromisifyAllOptions) {
```
callmehiphop: `typeof PubSub` might be more correct here?

test/index.ts
```diff
@@ -91,7 +101,7 @@ class FakeTopic {
 let extended = false;
 const fakePaginator = {
   // tslint:disable-next-line variable-name
-  extend(Class, methods) {
+  extend(Class: Function, methods: string[]) {
```
callmehiphop: `typeof PubSub` might be more correct here?

test/index.ts
```diff
@@ -137,13 +148,14 @@ defineOverridableClient('SubscriberClient');
...
+  let PubSub: typeof pubsubTypes.PubSub;
   const PROJECT_ID = 'test-project';
-  let pubsub;
+  // tslint:disable-next-line no-any
+  let pubsub: pubsubTypes.PubSub;
   const OPTIONS = {
     projectId: PROJECT_ID,
     promise: {},
-  };
+  } as pubsubTypes.ClientConfig;
```
callmehiphop: nitpick/question: why use `as` instead of

```js
const OPTIONS: pubsubTypes.ClientConfig = {
  // ...
};
```

test/index.ts
```diff
@@ -559,58 +564,67 @@ describe('PubSub', () => {
...
+          return topicInstance as pubsubTypes.Topic;
         };
 
-        pubsub.createTopic(topicName, (err, topic) => {
-          assert.ifError(err);
-          assert.strictEqual(topic, topicInstance);
-          done();
-        });
+        pubsub.createTopic(
+            topicName, (err?: Error|null, topic?: pubsubTypes.Topic|null) => {
```
callmehiphop: I'm surprised we need this, shouldn't the types be inferred?

test/index.ts
praveenqlogic: in that way giving error for promise member "Type '{}' is missing the following properties from type 'PromiseConstructor'"

test/index.ts
praveenqlogic: removed

test/index.ts
praveenqlogic: agree, here no need of declaring type, it is by default. 
image
in this way its look more informative

test/index.ts
praveenqlogic: Topic is a required parameter
https://github.com/googleapis/nodejs-pubsub/blob/master/src/index.ts#L339

test/index.ts
praveenqlogic: due to explicit type declaration give error "Argument of type 'undefined' is not assignable to parameter of type 'CreateSubscriptionOptions'"  (test looks valid with any type declaration for pubsub)

test/index.ts
praveenqlogic: fix suggestion by VS code, "infer type from usage"

src/topic.ts
```diff
@@ -376,13 +370,7 @@ export class Topic {
         callback!(null, this, apiResponse!);
         return;
       }
-      let code = 0;
```
JustinBeckwith: @callmehiphop can you think of a reason why this was done that way?

test/index.ts
JustinBeckwith: In this case we generally prefer to use `any[]`.  Could you remove this and do that instead?

test/index.ts
JustinBeckwith: Yes, TypeScript will get angry.  However, not all of our users are using TypeScript :) You need to cast pubsub as an `any` or cast `createSubsscription` as `Function` below.  Lets keep these tests.

test/index.ts
JustinBeckwith: Same deal as above.  Lets not use TypeScript as a reason to delete test coverage.  Cast to any as needed :)

test/index.ts
JustinBeckwith: No, please don't include the types here unless you have to.  Let the compiler do the work!

test/index.ts
```diff
@@ -1287,27 +1330,17 @@ describe('PubSub', () => {
...
-      subscriptionOverride = function(pubsub, name, options) {
+      subscriptionOverride = function(
+          pubsub: pubsubTypes.PubSub, name: string,
+          options: pubsubTypes.CreateSubscriptionOptions) {
         assert.strictEqual(options, CONFIG);
         done();
       };
       pubsub.subscription(SUB_NAME, CONFIG);
     });
-
```
JustinBeckwith: Please put these back

src/topic.ts
callmehiphop: No idea to be honest, looks like it got changed when no implicit any was turned on for this file. My best guess is that maybe `err` wasn't being inferred as a `ServiceError` so this was a work around to read the code?

JustinBeckwith: Who the hell knows.  Either way, fine to delete.

test/index.ts
praveenqlogic: done

test/index.ts
```diff
@@ -1248,16 +1283,10 @@ describe('PubSub', () => {
   });
 
   describe('snapshot', () => {
-    it('should throw if a name is not provided', () => {
```
JustinBeckwith: Can we add this back please?

test/index.ts
praveenqlogic: done,

JustinBeckwith:
@praveenqlogic apologies, few merge conflicts here now.  Can you merge and update?

praveenqlogic:
@JustinBeckwith : Hi merge conflict get resolved. let me know if any change needed.

PR #503, opened 2019-02-27
feat(message): use precise-date for message publish time

callmehiphop:
`publishTime` is returned with microsecond precision, seems like a great use case for precise date!

- [x] Tests and linter pass

### DIFF
--- a/src/subscriber.ts
+++ b/src/subscriber.ts
@@ -66,7 +66,7 @@ export interface PullResponse {
  *   //   data: Buffer.from('Hello, world!),
  *   //   id: '1551297743043',
  *   //   publishTime: new PreciseDate('2019-02-27T20:02:19.029534186Z'),
- *   //   received: 1551297743043
+ *   //   received: 1551297743043,
  *   //   length: 13
  *   // }
  * });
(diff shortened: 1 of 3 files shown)


src/subscriber.ts
```diff
@@ -46,33 +47,91 @@ export interface PullResponse {
...
- * @param {object} message The raw message response.
+ * @example
+ * subscription.on('message', message => {
+ *   // {
+ *   //   ackId: 'RUFeQBJMJAxESVMrQwsqWBFOBCEhPjA',
+ *   //   attributes: {key: 'value'},
+ *   //   data: Buffer.from('Hello, world!),
+ *   //   id: '1551297743043',
+ *   //   publishTime: new PreciseDate('2019-02-27T20:02:19.029534186Z'),
+ *   //   received: 1551297743043
```
stephenplusplus: Really silly, but this line in a comment is missing a trailing comma.

callmehiphop: Oof, nice catch!

callmehiphop:
@JustinBeckwith technically yes, but I don't expect it to impact anyone. (famous last words)

PR #513, opened 2019-03-02
fix: improve types and remove a few packages

JustinBeckwith:
- [x] Tests and linter pass

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -26,8 +26,7 @@ import {replaceProjectIdToken} from '@google-cloud/projectify';
 import {promisifyAll} from '@google-cloud/promisify';
 import * as extend from 'extend';
 import {GoogleAuth} from 'google-auth-library';
-import * as gax from 'google-gax';
-import * as is from 'is';
+import {ClientStub, GrpcClient} from 'google-gax';
 
 const PKG = require('../../package.json');
 const v1 = require('./v1');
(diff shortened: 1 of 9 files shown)


src/index.ts
```diff
@@ -87,6 +85,7 @@ export interface GetSubscriptionsCallback {
 }
 
 export interface GetTopicsOptions extends Options {}
+export type GetTopicsResponse = [Topic[], object];
```
callmehiphop: I'm actually working a PR that overhauls some of the types, but I think the signature for all the lists is incorrect. It can either be one of the following

```typescript
type AllTopics = [Topic[]];
type TopicsPage = [Topic[], {}|null, google.pubsub.v1.IListTopicsResponse]
export type GetTopicsResponse = AllTopics|TopicsPage; 
```

src/topic.ts
```diff
@@ -656,7 +647,7 @@ export class Topic {
   publishJSON(
       json: object, attributesOrCallback?: Attributes|PublishCallback,
       callback?: PublishCallback): Promise<string>|void {
-    if (!is.object(json)) {
+    if (!json || typeof json !== 'object') {
```
callmehiphop: This kind of check makes me nervous. What are your feelings on using a `try/catch` to wrap the call to `JSON.stringify()` instead?

PR #514, opened 2019-03-02
fix: update TS types for `PubSub#subscription` to match 0.23+'s API

ide:
As of 0.23 and later, `PubSub#subscription` and `Topic#subscription` take an options object with `streamingOptions` instead of `maxConnections`. Also, the methods that take this options object never use `gaxOpts` nor `autoPaginate` -- these fields are unused in these code paths.

This commit updates the TS types so that the options object passed to `PubSub#subscription` and `Topic#subscription` (a) has the new `streamingOptions` field and (b) does not have the unused `gaxOpts` and `autoPaginate` fields. As part of this commit, some TS `as` type assertions are no longer needed.

There is one small breaking change in this commit -- the options passed to `Topic#subscription` are no longer mutated. Instead, a new options object (of a different type) is created. Since the options are no longer mutated, a small change to one of the unit tests was necessary.

Test plan: TS builds and all tests pass

- [x] Tests and linter pass

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -33,6 +33,7 @@ const PKG = require('../../package.json');
 const v1 = require('./v1');
 
 import {Snapshot} from './snapshot';
+import {SubscriberOptions} from './subscriber';
 import {Subscription, SubscriptionMetadataRaw} from './subscription';
 import {Topic, PublishOptions} from './topic';
 import {CallOptions} from 'google-gax';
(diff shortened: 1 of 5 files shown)


callmehiphop:
@ide thanks for the PR! There were a few other type issues as well but I think I managed to get them all this morning. Unfortunately I have ruined your PR in the process :( I think I managed to cover everything here, but if you find anything else please let us know. Thanks again!

ide:
The new types look like they work for me, thanks!

PR #517, opened 2019-03-02
refactor: clean up unused packages

JustinBeckwith:

- [ ] Tests and linter pass


system-test/pubsub.ts
```diff
@@ -610,15 +596,10 @@ describe('pubsub', () => {
...
-      const snapshots: any[] = [];
-
+      const snapshots = new Array<Snapshot>();
       pubsub.getSnapshotsStream()
           .on('error', done)
-          .on('data',
-              snapshot => {
-                snapshots.push(snapshot);
-              })
+          .on('data', snapshots.push)
```
callmehiphop: I think you have to do it the less fun way D:

PR #518, opened 2019-03-03
Release @google-cloud/pubsub v0.26.1

JustinBeckwith:
This pull request was generated using releasetool.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,25 @@
 
 [1]: https://www.npmjs.com/package/nodejs-pubsub?activeTab=versions
 
+## v0.26.1
+
+03-03-2019 11:42 PST
+
+### Bug Fixes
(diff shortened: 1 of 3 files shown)


callmehiphop:
@JustinBeckwith how do version TypeScript changes? Part of my PR removed a a duplicate interface with a different name. I know a number of the definitions were incorrect, but do we treat something like that as a patch or breaking change?

JustinBeckwith:
You know... I don't know what the right answer is here.  @ofrobots and the @googleapis/node-team - any guidance on how to treat breaking TypeScript changes with regards to semver?

JustinBeckwith:
@callmehiphop yeah - you're right.  This should be a breaking change, and a minor version bump.  We're gettin really good at those with this lib lol. Can you cut a 0.27 release today?

PR #522, opened 2019-03-05
refactor(typescript): noImplicitAny for message-stream test file

praveenqlogic:

- [ ] Tests and linter pass

### DIFF
--- a/test/message-stream.ts
+++ b/test/message-stream.ts
@@ -15,11 +15,13 @@
  */
 
 import * as assert from 'assert';
-import {Metadata} from 'grpc';
+import {Metadata, ServiceError} from 'grpc';
 import * as proxyquire from 'proxyquire';
 import * as sinon from 'sinon';
 import {Duplex, PassThrough} from 'stream';
 import * as uuid from 'uuid';
+import * as messageTypes from '../src/message-stream';
+import {Subscriber} from '../src/subscriber';
 
 // just need this for unit tests.. we have a ponyfill for destroy on
 // MessageStream and gax streams use Duplexify
@@ -112,7 +114,7 @@ class FakeSubscriber {
   name: string;
   ackDeadline: number;
   client: FakeGaxClient;
-  constructor(client) {
+  constructor(client: FakeGaxClient) {
     this.name = uuid.v4();
     this.ackDeadline = Math.floor(Math.random() * 600);
     this.client = client;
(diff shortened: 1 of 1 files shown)


test/message-stream.ts
```diff
@@ -142,7 +144,7 @@ describe('MessageStream', () => {
     const gaxClient = new FakeGaxClient();
     client = gaxClient.client;  // we hit the grpc client directly
     subscriber = new FakeSubscriber(gaxClient);
-    messageStream = new MessageStream(subscriber);
+    messageStream = new MessageStream(subscriber as {} as Subscriber);
```
callmehiphop: I'm seeing this cast all over the test file. Why not just cast the subscriber in the line above?

test/message-stream.ts
```diff
@@ -156,20 +158,21 @@ describe('MessageStream', () => {
         objectMode: true,
         highWaterMark: 0,
       };
-
-      assert.deepStrictEqual(messageStream.options, expectedOptions);
+      // tslint:disable-next-line no-any
+      assert.deepStrictEqual((messageStream as any).options, expectedOptions);
```
callmehiphop: If you cast this as `FakePassThrough` you can avoid the any cast/comment.

test/message-stream.ts
```diff
@@ -156,20 +158,21 @@ describe('MessageStream', () => {
...
+          new MessageStream(subscriber as {} as Subscriber, {highWaterMark});
 
       const expectedOptions = {
         objectMode: true,
         highWaterMark,
       };
-
-      assert.deepStrictEqual(ms.options, expectedOptions);
+      // tslint:disable-next-line no-any
+      assert.deepStrictEqual((ms as any).options, expectedOptions);
```
callmehiphop: If you cast this as `FakePassThrough` you can avoid the any cast/comment.

test/message-stream.ts
```diff
@@ -338,7 +344,7 @@ describe('MessageStream', () => {
         const fakeResponses = [{}, {}, {}, {}, {}];
         const recieved: object[] = [];
 
-        messageStream.on('data', chunk => recieved.push(chunk))
+        messageStream.on('data', chunk => recieved.push(chunk as object))
```
callmehiphop: Seems like if we can avoid casting emitted data we should. Does implementing an interface that lists the correct types work here?

```typescript
interface StreamEvents {
  on(event: 'data', listener: (data: {}) => void): void;
}

class MessageStream implements StreamEvents {}
```

?

test/message-stream.ts
praveenqlogic: Done

test/message-stream.ts
praveenqlogic: done thanks 👍

PR #525, opened 2019-03-05
fix(typescript): correct response type of `Subscription.get`

### LINKED ISSUE #507: Invalid typings
#### Environment details

  - OS: irrelevant
  - Node.js version: irrelevant
  - npm version: irrelevant
  - `@google-cloud/pubsub` version: 0.26.0

It looks like the typings shipped with 0.26.0 are incorrect. I can't see any difference in actual implementation nor the changelog that would explain these changes:
- `Topic.exists` has a required parameter - a callback. It should return void if provided with a callback, a promise otherwise.
- `Topic.get` returns a `Topic`, not `[Topic]` as it used to. Although I'd gladly welcome this change to the interface, it seems that it actually didn't change and still returns an array.

There might be more, I just pointed out the ones I've came across and tested.

merlinnot:
Resolves #507 

- [x] Tests and linter pass


callmehiphop:
Looks like we're getting a small lint error 

```
src/topic.ts:181:49 - error TS2345: Argument of type 'ResourceCallback<Topic, ITopic> | undefined' is not assignable to parameter of type 'ResourceCallback<Topic, ITopic>'.
  Type 'undefined' is not assignable to type 'ResourceCallback<Topic, ITopic>'.

181     this.pubsub.createTopic(this.name, gaxOpts, callback);
```

I think we just need to make the following edit

```js
this.pubsub.createTopic(this.name, gaxOpts, callback!);
```

merlinnot:
The call relied on an internal implementation (checking if the argument is equal to `undefined`), so I made it more correct so it actually conforms to a function interface.

callmehiphop:
@merlinnot we actually use a decorator to promisify everything, so `callback` will never be undefined, however we've been marking it as optional for the Promise signature overload.

merlinnot:
Ok, changed and rebased on master.

callmehiphop:
@merlinnot awesome, thank you so much!

merlinnot:
I can't see CI logs, both `lint` and `compile` work for me locally. It's hard to contribute when only Googlers see build logs ;)

callmehiphop:
@merlinnot yeah T_T I don't think there's anything wrong with your code, Kokoro (CI) has been super flakey today. @JustinBeckwith do we have any idea what's going on with it?

PR #528, opened 2019-03-06
refactor: type fixes and some light housekeeping

### LINKED ISSUE #507: Invalid typings

### LINKED ISSUE #526: I just tried to upgrade and I am getting many cannot find type "Long" error

#### Environment details

  - OS: Mac OS X
  - Node.js version: node v10
  - npm version: yarn 1.13.0
  - `@google-cloud/pubsub` version: Current 0.24.1  Tried to Upgrade to 0.27.0

#### Steps to reproduce

  1. Setup existing project using typescript
  2. Upgrade from 0.24.1 to 0.27.0

I don't know if there are other causes or setups I need to do. I have tried to add "types: ['Long']" into the compiler configuration but that didn't help. I found references to some other projects also using protobuf with similar problems, but I was unable to find a place where it presented a good solution.
I tried quite a few different approaches, but didn't get anywhere, sorry.

callmehiphop:
Fixes #507
Fixes #526
Closes #525

- [x] Tests and linter pass


test/snapshot.ts
```diff
@@ -233,14 +233,5 @@ describe('Snapshot', () => {
 
       snapshot.delete(done);
     });
-
```
JustinBeckwith: So...  what's up with the deleted test?

callmehiphop: They were useless. We would do a `callback = callback || noop` but callback will never be undefined because omitting it just users the promisified version.

JustinBeckwith:
LGTM, assuming tests pass :)

PR #534, opened 2019-03-08
refactor(typescript): noImplicitAny for Subscription test file

### LINKED ISSUE #313: Enable noImplicitAny in the tsconfig

praveenqlogic:
Fixes #313 (it's a good idea to open an issue first for discussion)

All files in test, src and system-test folder is done!

- [ ] Tests and linter pass

### DIFF
--- a/test/subscription.ts
+++ b/test/subscription.ts
@@ -17,27 +17,26 @@
 import * as pfy from '@google-cloud/promisify';
 import * as assert from 'assert';
 import {EventEmitter} from 'events';
-import {CallOptions} from 'google-gax';
 import {ServiceError} from 'grpc';
 import * as proxyquire from 'proxyquire';
 import * as sinon from 'sinon';
 
-import {google} from '../proto/pubsub';
-import {ExistsCallback, RequestCallback, RequestConfig} from '../src/pubsub';
-import {SeekCallback, Snapshot} from '../src/snapshot';
+import {PubSub, RequestConfig} from '../src/pubsub';
+import {Snapshot} from '../src/snapshot';
 import {Message, SubscriberOptions} from '../src/subscriber';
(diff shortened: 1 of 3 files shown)


test/subscription.ts
```diff
@@ -596,123 +576,103 @@ describe('Subscription', () => {
...
               callback(null);  // the done fn
-            };
+            });
 
         subscription.get(options, done);
       });
     });
 
     describe('error', () => {
+      afterEach(() => sandbox.restore());
```
JustinBeckwith: I'd prefer we pull this up way to the top, and do this in `afterEach` for ever test in the file.

test/subscription.ts
```diff
@@ -596,123 +576,103 @@ describe('Subscription', () => {
...
       it('should pass back errors when not auto-creating', done => {
         const error = {code: 4} as ServiceError;
         const apiResponse = {};
-
-        subscription.getMetadata =
-            (gaxOpts: CallOptions,
-             callback: GetSubscriptionMetadataCallback) => {
+        sandbox.stub(subscription, 'getMetadata')
```
JustinBeckwith: Instead of this you can do:
```js
sandbox.stub(subscription, 'getMetadata').callsArgWith(1, error, apiResponse);
```

test/subscription.ts
callmehiphop: I think this already exists in a global hook

praveenqlogic: Done

test/subscription.ts
praveenqlogic: Done

Issue #496, opened 2019-02-22
system tests leftovers


alexander-fenster:
Every system test run leaves behind 3 topics, 5 subscriptions, and 1 snapshot. Seems like we either don't delete something, or delete incorrectly. Within time, it all affects test runtime and, even worse, smoke tests start spamming a lot since they print each topic and each subscription in the project.

Could we make sure we cleanup after system tests properly?

PR #537, opened 2019-03-08
feat(topic): create setMetadata method

callmehiphop:
Relates to #496

- [x] Tests and linter pass


src/topic.ts
```diff
@@ -656,6 +664,78 @@ export class Topic {
...
+   */
+  /**
+   * @callback SetTopicMetadataCallback
+   * @param {?Error} err Request error, if any.
+   * @param {object} apiResponse The full API response.
+   */
+  /**
+   * Updates the topic.
+   *
+   * @see [UpdateTopicRequest API Documentation]{@link https://cloud.google.com/pubsub/docs/reference/rest/v1/UpdateTopicRequest}
```
bcoe: from reading this documentation, am I reading correctly that labels are the only "meta information" you would potentially change on a topic, and this method since it's calling `updateTopic` would also potentially be able to change the `name` of a topic?

I wonder if a better method name would just be `updateTopic`? which would better match the upstream API too?

stephenplusplus:
Why isn't Topic a ServiceObject?

JustinBeckwith:
So I had that question at first too.  Then I noticed - this module doesn't use any of the `common` modules 😆

src/topic.ts
```diff
@@ -38,8 +40,14 @@ export type GetTopicResponse = TopicResponse;
 
 export type GetTopicOptions = CallOptions&{autoCreate?: boolean};
 
-export type GetTopicMetadataCallback = RequestCallback<google.pubsub.v1.ITopic>;
-export type GetTopicMetadataResponse = [google.pubsub.v1.ITopic];
+type MetadataCallback = RequestCallback<TopicMetadata>;
+type MetadataResponse = [TopicMetadata];
+
+export type GetTopicMetadataCallback = MetadataCallback;
```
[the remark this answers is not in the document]
callmehiphop: At this point it would be another breaking change, I thought maybe I'd give ya'll a break from those :p

src/topic.ts
callmehiphop: `setMetadata` is the generic method name we use across all the libraries to update resource fields. I'm definitely not married to it or anything, just went with it for the sake of consistency. I'd be ok with moving away from some of our older conventions to better mirror the upstream apis.

@JustinBeckwith @stephenplusplus @jkwlui what do you guys think?

callmehiphop:
@stephenplusplus My memory is kind of fuzzy on this, but I think when we switched from the rest api to grpc the direction was to just use gax and skip all the common lib stuff.

PR #541, opened 2019-03-11
fix(typescript): correctly import long

### LINKED ISSUE #526: I just tried to upgrade and I am getting many cannot find type "Long" error

callmehiphop:
Fixes #526

- [x] Tests and linter pass


package.json
```diff
@@ -44,10 +44,9 @@
     "compile": "tsc -p . && cp -r src/v1 build/src/ && cp -r protos build/ && cp -r proto build/ && cp smoke-test/*.js build/system-test && cp test/*.js build/test",
     "prepare": "npm run compile",
     "pretest": "npm run compile",
-    "proto": "npm run proto:pubsub",
-    "proto:pubsub": "pbjs -t static-module -w commonjs -p protos google/pubsub/v1/pubsub.proto | pbts -i long -o proto/pubsub.d.ts -",
-    "proto:iam": "pbjs -t static-module -w commonjs -p protos google/iam/v1/iam_policy.proto | pbts -i long -o proto/iam.d.ts -",
-    "proto-types": "mkdir -p proto && npm run proto:pubsub && npm run proto:iam",
+    "proto": "mkdir -p proto && npm run proto:pubsub && npm run proto:iam",
```
bcoe: this script updates seem unrelated; was this making local testing hard because it would fail if you hadn't already created the proto filter?

I might suggest pulling this into another change, and instead adding the `mkdirp` dependency as a developer dependency, as `mkdir -p` won't work on windows.

callmehiphop: We had 2 (almost) identical scripts, I had created one of them so I figured we might as well dump it, but if you'd prefer it as a separate commit I'm cool to make another pr!

bcoe: If it's not too much of a pain, let's do another pr with `mkdirp` rather than the OS `mkdir -p` if you don't mind? this habit just better future-proofs us for Windows.

callmehiphop: Done! PTAL!

PR #542, opened 2019-03-11
refactor: fix/simplify proto gen scripts

callmehiphop:
Per https://github.com/googleapis/nodejs-pubsub/pull/541, this PR cleans up the proto scripts and uses `mkdirp` for Windows support!


callmehiphop:
@bcoe we do some Windows testing, but our current CI setup isn't very friendly to non-Googlers, so it's a little bit of a mystery to me to be honest, haha. @jkwlui could probably share some details with us!

PR #548, opened 2019-03-12
WIP: Add job to publish docs with docuploader

busunkim96:


.kokoro/publish-docs.sh
```diff
@@ -0,0 +1,38 @@
...
+npm install
+
+npm run docs
+
+# Publish documentation with docuploader.
+python3.6 -m pip install gcp-docuploader
+
+python3.6 -m docuploader create-metadata \
+			--name pubsub \
+			--version 0.27.0 \
```
JustinBeckwith: Is this the version of the npm module?  That feels like a scary thing to have hard-coded in here.

.kokoro/publish-docs.sh
```diff
@@ -0,0 +1,38 @@
...
+			--language node \
+			--distribution-name @google-cloud/pubsub \
+			--product-page https://cloud.google.com/pubsub/docs/ \
+			--github-repository https://github.com/googleapis/nodejs-pubsub \
+			--issue-tracker https://github.com/googleapis/nodejs-pubsub/issues
+			docs/docs.metadata 
+
+python3.6 -m docuploader upload docs
```
JustinBeckwith: Can we add a newline just so we don't have to see the weird little red thingy?

.kokoro/release/publish-docs.cfg
```diff
@@ -0,0 +1,34 @@
...
+# Download trampoline resources.
+gfile_resources: "/bigstore/cloud-devrel-kokoro-resources/trampoline"
+
+# Use the trampoline script to run in docker.
+build_file: "nodejs-pubsub/.kokoro/trampoline.sh"
+
+# Configure the docker image for kokoro-trampoline.
+env_vars: {
+    key: "TRAMPOLINE_IMAGE"
+    value: "gcr.io/cloud-devrel-kokoro-resources/node:8-user"
```
JustinBeckwith: Can we use node:10 here?

.kokoro/release/publish-docs.cfg
```diff
@@ -0,0 +1,34 @@
...
+}
+
+env_vars: {
+    key: "TRAMPOLINE_BUILD_FILE"
+    value: "github/nodejs-pubsub/.kokoro/publish-docs.sh"
```
JustinBeckwith: Silly question - what actually calls this script?  I would assume it's part of the release publish script, no?

.kokoro/publish-docs.sh
busunkim96: Sorry for the scare, I'm using this branch to iron out any weirdness before creating a PR to update synthtool templates. I will fetch the version (and the name) from the appropriate place. What's the best way to get the npm module version?

.kokoro/release/publish-docs.cfg
busunkim96: I was planning on chaining this job (publish-docs) to the release publish job. Would it make more sense for the package publishing and docs publishing to happen in one job?

JustinBeckwith: Nah, whatever you have planned is fine.  Was just curious :)

.kokoro/publish-docs.sh
JustinBeckwith: Assuming the version has been published to npm - you can run:
```
npm view @google-cloud/pubsub version
```

otherwise, I think you'd have to parse `package.json`, and read the `version` property.  @bcoe do you know if there's a way to fetch the current version from `package.json` with the `npm` cli?

bcoe: I've tended to do something like this if I'm getting really creative:

`curl -XGET https://registry.npmjs.org/@google-cloud%2Fpubsub | json versions.'["0.28.1"]'`

I don't know off the top of my head if there is a way to do this with the `npm` bin; worth noting that a lot of npm commands accept `--json`, so if you do find a command that works you can use this to output an easily parsed value -- I love the tool [json](https://www.npmjs.com/package/json) for then facilitating transformations on the JSON data.

PR #556, opened 2019-03-20
feat(subscription): accept pull deadline option

callmehiphop:
- [x] Tests and linter pass

### DIFF
--- a/src/message-stream.ts
+++ b/src/message-stream.ts
@@ -53,11 +53,19 @@ const RETRY_CODES: status[] = [
 ];
 
 /*!
+ * Deadline for the stream.
+ */
+const PULL_DEADLINE = require('./v1/subscriber_client_config.json')
+                          .interfaces['google.pubsub.v1.Subscriber']
+                          .methods.StreamingPull.timeout_millis;
+
+/*!
  * default stream options
  */
 const DEFAULT_OPTIONS: MessageStreamOptions = {
   highWaterMark: 0,
   maxStreams: 5,
+  pullDeadline: PULL_DEADLINE,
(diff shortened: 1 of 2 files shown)


src/message-stream.ts
```diff
@@ -115,11 +123,15 @@ export class ChannelError extends Error implements ServiceError {
  *     {@link https://nodejs.org/en/docs/guides/backpressuring-in-streams/} for
  *     more details.
  * @property {number} [maxStreams=5] Number of streaming connections to make.
+ * @property {number} [pullDeadline=900000] Deadline to be applied to each
```
stephenplusplus: The internal "deadline" that we use through gapic wants a time valur, but we repurpose "deadline" here to mean the amount of time from now. Should we just have the user provide a time value as well, or maybe rename this setting to "streamAliveMs" (or something)?

callmehiphop: We actually bypass the gapic for this class and use plain old grpc. Definitely not partial to the name though

stephenplusplus: Ah, ok. But I think the naming logic is still the same, where deadline is a time stamp in gRPC land, whereas that's not how the new option would be used. I think the name should not include "deadline" only for that reason. As far as other names, I'm open to ideas as well!

callmehiphop: sure, `pullTimeout` then?

stephenplusplus: That makes sense to me!

callmehiphop: Done. PTAL!

stephenplusplus: Thank you sir!

PR #560, opened 2019-03-25
feat(subscription): ordered messages

callmehiphop:
This will add support for `orderingKey` in the handwritten layer.

- [x] Tests and linter pass

### DIFF
--- a/src/subscriber.ts
+++ b/src/subscriber.ts
@@ -21,31 +21,16 @@ import {EventEmitter} from 'events';
 import {ClientStub} from 'google-gax';
 import {common as protobuf} from 'protobufjs';
 
+import {google} from '../proto/pubsub';
+
 import {Histogram} from './histogram';
 import {FlowControlOptions, LeaseManager} from './lease-manager';
 import {AckQueue, BatchOptions, ModAckQueue} from './message-queues';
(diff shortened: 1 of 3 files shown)


src/subscriber.ts
```diff
@@ -94,36 +81,51 @@ export class Message {
      * @name Message#ackId
      * @type {string}
      */
-    this.ackId = ackId;
+    this.ackId = ackId!;
```
JustinBeckwith: Why is the `!` needed here?

src/subscriber.ts
```diff
@@ -94,36 +81,51 @@ export class Message {
      * @name Message#ackId
```
JustinBeckwith: For the constructor, can we use a real interface instead of the inline type?

JustinBeckwith (changes requested):

src/subscriber.ts
callmehiphop: I switched to use the generated types, before I just hand wrote it but I don't think we had a solution in place for that yet. As such, they mark everything as optional (even though that isn't necessarily the case)

src/subscriber.ts
callmehiphop: Sure!

JustinBeckwith (approved):

PR #566, opened 2019-04-05
fix: relax strictEqual to match RegExp

jkwlui:
automerge merged the last PR before my changes were pushed.

### DIFF
--- a/samples/system-test/subscriptions.test.js
+++ b/samples/system-test/subscriptions.test.js
@@ -58,9 +58,9 @@ describe('subscriptions', () => {
     const output = execSync(
       `${cmd} create ${topicNameOne} ${subscriptionNameOne}`
     );
-    assert.match(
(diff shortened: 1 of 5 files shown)


samples/system-test/subscriptions.test.js
```diff
@@ -67,7 +70,10 @@ describe('subscriptions', () => {
     const output = execSync(
       `${cmd} create-push ${topicNameOne} ${subscriptionNameTwo}`
     );
-    assert.strictEqual(output, `Subscription ${subscriptionNameTwo} created.`);
+    assert.match(
```
bcoe: I'd be tempted to just use `assert.include` (which we discussed in chat), but the match works too :+1: 

In the tests I've been fixing up I've been using `include` rather than `strictEqual` and `notInclude` rather than `strictNotEqual`.

PR #567, opened 2019-04-06
chore: switch to include from match

bcoe:
checking to see if the problem with tests was related to invalid regexes being created

- [x] Tests and linter pass


bcoe:
@jkwlui just a hunch, might not do the trick; it was feeling like creating `RegExp` was also a lot of ritual.

PR #569, opened 2019-04-09
Clarify time units for nack delay arg

ncjones:
Reading the docs I wasn't sure whether nack(delay) should be provided seconds or millis. I assume the docs are generated from these sources.


sduskis:
@ncjones, we can't do anything on this pull request until you sign the CLA.

PR #576, opened 2019-04-18
feat(subscription): allow for message polling

### LINKED ISSUE #468: All acks are expired after 30 min of activity with v0.23.0+
#### Environment details
  - Hardware: on-prem 8-core compute node
  - OS: RHEL 6.9
  - Node.js version: 8.11.3
  - npm version: 6.4.0
  - `@google-cloud/pubsub` 0.23.0-0.24.1:

#### Steps to reproduce

  1. Publish a moderate number of messages to a topic (~4k in our case)
  2. Use a single subscriber that processes each message within 1-2 min, e.g. with the following code:
```ts
const subscription = pubSub.subscription(SUBSCRIPTION_NAME, {
  flowControl: {
    maxMessages: 2,
    allowExcessMessages: false,
  },
});
subscription.on("message", ({ ack }) => {
  setTimeout(ack, 60000 * 2 * Math.random());
});
subscription.on("error", console.error);
``` 
  3. After ~30 min of churning through messages, Stackdriver shows that:
    - `num_undelivered_messages` stops going down
    - `pull_ack_message_operation_count` marks every message ack as `expired`
        screen shot 2019-02-07 at 2 11 10 am
    - `streaming_pull_message_operation_count` shows that 10-min pull interval stops.
        screen shot 2019-02-07 at 1 53 23 am

As a net result, we can't seem to ack any more messages and have to restart the subscriber.

We varied settings like `flowControl.maxMessages` (1-4) and `ackDeadline` (10, 300, 7200, 720000) and that didn't seem to have an effect.

Downgrading to 0.22.2 seems to resolve this issue. We still get occasional `expired` acks, but not in 100% of cases.

Thanks!

callmehiphop:
Fixes #468

This allows for users to configure a subscription to use polling strategy via `pull` requests. It will be disabled by default, but users who don't want/need high throughput might consider using this in an effort to cut back on duplicate messages.

- [ ] Tests and linter pass


callmehiphop:
@rossj if you wouldn't mind testing this PR out that would be awesome, just need to pass in a `poll` option to enable it.

```js
const sub = pubsub.subscription('my-sub', {poll: true});
```

src/message-stream.ts
```diff
@@ -384,3 +382,104 @@ export class MessageStream extends PassThrough {
     }
   }
 }
+
+/**
```
bcoe: how painful would it be to pull this into its own file? I'd be tempted to have a `message-stream.ts` and a `polling-message-stream.js`.

callmehiphop: It would be super easy! SGTM

callmehiphop:
> Is it expected behavior that a long-lived connection will receive duplicate messages when throughput is high?

@bcoe there are a number of factors that can cause an increase in duplicate messages. In regards to #468 gRPC is internally buffering messages and if a user doesn't ack them fast enough, by the time they make it to the client they're already expired and likely scheduled to be redelivered. Ideally we would fine tune the underlying flow control, but as far as I know that is not possible

src/polling-message-stream.ts
```diff
@@ -62,28 +64,18 @@ export class PollingMessageStream extends Readable {
     this._reading = false;
   }
   /**
-   * Destroys the stream and any underlying streams.
+   * Cancels any active requests once the stream is destroyed.
    *
-   * @param {error?} err An error to emit, if any.
    * @private
    */
-  destroy(err?: Error): void {
```
callmehiphop: With v6 reaching EOL in a few days, I decided to just remove this and wait to merge until we've dropped support for it.

callmehiphop:
@bcoe since you've last reviewed I've written some tests and did a couple minor refactors. Would be great if you could take another glance when you get a chance

test/polling-message-stream.ts
```diff
@@ -0,0 +1,241 @@
...
+    subscriber = (new FakeSubscriber(client) as {}) as Subscriber;
+    stream = new PollingMessageStream(subscriber, OPTIONS);
+  });
+
+  afterEach(() => {
+    stream.destroy();
+    sandbox.restore();
+  });
+
+  describe('instantiation', () => {
```
bcoe: really appreciate the additional tests 👍

src/polling-message-stream.ts
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * @typedef {object} MessagePollingOptions
+ * @property {number} [batchSize=100] Desired number of messages per batch.
+ *     Defaults to {@link FlowControlOptions} `maxMessages` value (100).
+ * @property {number} [highWaterMark=0] Configures the Buffer level. See
+ *     {@link https://nodejs.org/en/docs/guides/backpressuring-in-streams/} for
+ *     more details.
+ */
+export interface MessagePollingOptions {
```
bcoe: appreciate that you broke this into its own class.

callmehiphop:
I'm not really sure where we stand with this, I'm going to put it on hold until I get further direction.

callmehiphop:
FYI @sduskis in case you want to take a look in regards to https://github.com/googleapis/nodejs-pubsub/issues/468#issuecomment-493612945

sduskis:
Please don't merge this.  We need to work with the Pub/Sub team to see how to roll out this feature across all languages.

callmehiphop:
Hasn't been any movement here in a while, so I'm going to close this for now.

PR #590, opened 2019-05-03
DO NOT MERGE: refactor: use types from @grpc/grpc-js

alexander-fenster:
DO NOT MERGE until gax v1.0.0 (with `@grpc/grpc-js` by default) is shipped.

This pull request removes all references to `grpc` in types and fixes some type incompatibiity errors (mostly related to `Error` types).

- [x] Tests and linter pass
  - note: system test 📦 does not work because of outdated `@grpc/grpc-js`, will be fixed within gax release.


alexander-fenster:
@callmehiphop Dave, this should go in after I release gax, but before we release Pub/Sub :) Could you please take a look and let me know if anything looks weird?

Thanks!

src/message-stream.ts
```diff
@@ -117,9 +119,13 @@ export class StatusError extends Error implements ServiceError {
  */
 export class ChannelError extends Error implements ServiceError {
   code: status;
+  metadata: Metadata;
+  details: string;
   constructor(err: Error) {
     super(`Failed to connect to channel. Reason: ${err.message}`);
     this.code = err.message.includes('deadline') ? DEADLINE : UNKNOWN;
+    this.metadata = new Metadata();
+    this.details = '';
```
callmehiphop: Did `details` become a required property?

src/pubsub.ts
```diff
@@ -63,7 +68,11 @@ export interface ClientConfig extends gax.GrpcClientOptions {
   apiEndpoint?: string;
   servicePath?: string;
   port?: string | number;
-  sslCreds?: ChannelCredentials;
+  // TODO: sslCreds is supposed to be of type ChannelCredentials.
+  // Using any until google-gax changes to @grpc/grpc-js by default.
+  // tslint:disable-next-line no-any
+  sslCreds?: any;
```
callmehiphop: Can we open a tracking issue for this?

src/pubsub.ts
```diff
@@ -897,7 +906,12 @@ export class PubSub {
     if (!hasProjectId && !this.isEmulator) {
       this.auth.getProjectId((err, projectId) => {
         if (err) {
-          callback(err);
+          const serviceError = Object.assign(err, {
```
callmehiphop: Just a nit, but maybe we should create a custom `UnknownError` for instances like this?

src/message-stream.ts
alexander-fenster: Yes, I believe so. https://github.com/grpc/grpc-node/blob/f7f098b1173cc4d58da0ef756df097df3fc10c6a/packages/grpc-js/src/call-stream.ts#L46

src/pubsub.ts
alexander-fenster: Done! https://github.com/googleapis/nodejs-pubsub/issues/595

callmehiphop:
@alexander-fenster LGTM! 👍

alexander-fenster:
@callmehiphop I cannot merge it right now because the latest gax release depends on grpc-js ^0.3.0 and the packaging test won't work until we start using v0.4.0. What I think I will do is release gax v1.0.0 with grpc-js only (and v0.4.0) later this week, then update this PR to use gax v1.0.0 (the CI should work then) and merge it. So let this wait for a while.

As for Windows build failure, this is being taken care of (PR is coming).

callmehiphop:
@alexander-fenster it sounds like we don't plan on switching to grpc-js any time too soon. I think it might make sense to hold off on using these types. WDYT?

alexander-fenster:
Yes, let's hold this for now.

Issue #318, opened 2018-11-05
Subscription does not recover from authentication credentials error (token expiry?)


eug48:
#### Environment details

  - OS: Linux, GKE container
  - Node.js version: 8.12.0
  - npm version: 6.4.1
  - `@google-cloud/pubsub` version: 0.20.1

#### Steps to reproduce

  1. Created a subscription
```js
import PubSub = require('@google-cloud/pubsub');
const gcpPubsubClient = new PubSub({ projectId });
const gcpPubsubSubscription = gcpPubsubClient.subscription(subscriptionId);
gcpPubsubSubscription.on('message', pubsubMsg => { ... })
gcpPubsubSubscription.on('error', err => { log(...) })
```
  2. Subscription works fine for many hours
  3. Suddenly in the middle of the night I get an uncaughtException:
```
uncaughtException { Error: 16 UNAUTHENTICATED: Request had invalid authentication credentials. Expected OAuth 2 access token, login cookie or other valid authentication credential. See https://developers.google.com/identity/sign-in/web/devconsole-project.
    at Object.exports.createStatusError (/app/node_modules/grpc/src/common.js:87:15)
    at Object.onReceiveStatus (/app/node_modules/grpc/src/client_interceptors.js:1188:28)
    at InterceptingListener._callNext (/app/node_modules/grpc/src/client_interceptors.js:564:42)
    at InterceptingListener.onReceiveStatus (/app/node_modules/grpc/src/client_interceptors.js:614:8)
    at callback (/app/node_modules/grpc/src/client_interceptors.js:841:24)
```

Bunyan also logged some metadata:
```
 error: {
   code: 16    
   details: "Request had invalid authentication credentials. Expected OAuth 2 access token, login cookie or other valid authentication credential. See https://developers.google.com/identity/sign-in/web/devconsole-project."    
   
metadata: {
_internal_repr: {  
www-authenticate: [
      0: "Bearer realm="https://accounts.google.com/", error="invalid_token""       
     ]
    }
   }
  }
```

The gcpPubsubSubscription `error` handler also fires:
```
err: {
   code: 16    
   message: "Request had invalid authentication credentials. Expected OAuth 2 access token, login cookie or other valid authentication credential. See https://developers.google.com/identity/sign-in/web/devconsole-project."    
   name: "Error"    
[...]

PR #594, opened 2019-05-06
Add max retry while creating streams in subscription

### LINKED ISSUE #550: Many parallel subscriptions in one process => DEADLINE_EXCEEDED upon ack
#### Environment details

  - OS: GKE
  - Node.js version: 10.13.0
  - `@google-cloud/pubsub` version: 0.28.1

#### Steps to reproduce

This is a follow-up report to #240, the symptoms are the same however I think I have a better understanding of what might go wrong this time.

I have made another take on finally upgrading the node pubsub library from `v0.18.0` to `v0.28.1`.

* I am currently running a couple of pods that subscribe to several pubsub topics.
* Most of the pods subscribe to only one or two topics, those work fine with the updated library.
* I have one pod that subscribes to many subscriptions, this one shows problems, see below.

Here's what the problematic pod does:

* Starts up by initializing 23 different (pre-existing) subscriptions on 23 different topics.
* Starts pulling messages that sit in these subscriptions, these are low-frequency subscriptions, so we're talking about 0-500 messages that sit in the subscriptions initially.
* Processes those messages in batches and acks() all of them, processing of a batch takes ~500ms.
* After a few seconds (I've seen 5-12), I receive a `DEADLINE_EXCEEDED` error event on one of the subscriptions where acks have been sent (and I crash the process).

Here's the error:

```
Error: Failed to "acknowledge" for 6 message(s). Reason: 4 DEADLINE_EXCEEDED: Deadline Exceeded
    at AckQueue.<anonymous> (/app/node_modules/@google-cloud/pubsub/build/src/message-queues.js:162:23)
    at Generator.throw (<anonymous>)
    at rejected (/app/node_modules/@google-cloud/pubsub/build/src/message-queues.js:20:65)
```

For the affected subscriptions, the `num_undelivered_messages` stackdriver metric just keeps on rising:

Screenshot 2019-03-13 22 25 11

Stackdriver also indicates that there are indeed `streaming_pull_message_operation_count` operations (they keep on rising because my pod keeps on crashing and restarting, thus pulling the rising amount of unacked messages over and over again):

Screenshot 2019-03-13 22 26 44

However stackdriver doesn't show any `pull_ack_message_operation_count` operations (ignore the spike to the right, this is where I started the pod with only one subscription where it worked):

Screenshot 2019-03-13 22 30 30

Just for comparison, with version `v0.18.0` of this client library, that I had been using before, it was still using `streamingAcks` (see discussion in #240) and everything was working well, even with the 23 parallel subscriptions in one node process:

Screenshot 2019-03-13 22 32 18

All screenshots above show the same subscription and the same timeframe, at time ~4:30pm I deployed the new version, the point where the amount of undelivered messages starts rising.

To me it looks like the amount of parallel subscriptions determines whether acks work as expected or whether I consistently receive `DEADLINE_EXCEEDED` errors as a result to the `ack` batches. It's interesting that acks either work for all subscriptions handled by the affected process or none.

I doubt that it is CPU related because the process does basically only I/O (pipe the messages from Pubsub to BigQuery) and the CPU-graphs of GKE also show no significant CPU usage. CPU usage of the same process with `v0.18.0` of the library was also <1%.

Is the number of parallel subscriptions that can be handled in a single node process (all subscription message handlers registered with a single `new PubSub()` client instance) somehow limited?

Any pointer into what direction I should be looking to would be highly appreciated!

(Next thing I'd like to try out is see if I can narrow down the number of parallel subscriptions in one process where it starts to break… However due to this npm registry issue I have currently trouble deploying new code to the GKE cluster: https://status.npmjs.org/incidents/3p30y0pkblsp)

ajaaym:
Fixes #550 #318

- [ ] Tests and linter pass

### DIFF
--- a/src/message-stream.ts
+++ b/src/message-stream.ts
@@ -44,6 +44,16 @@ const DEADLINE: status = 4;
 const UNKNOWN: status = 2;
 
 /*!
+ * Max number of retries
+ */
+const DEFAULT_MAX_RETRIES = 3;
+
+/*!
+ * Time to reset retries counter.
+ */
+const RESET_RETRIES_COUNT_TIME_MS = 30000;
+
+/*!
  * codes to retry streams
  */
 const RETRY_CODES: status[] = [
@@ -56,6 +66,7 @@ const RETRY_CODES: status[] = [
   13, // internal error
   14, // unavailable
   15, // dataloss
+  16,  // unauthorized
 ];
 
 /*!
@@ -72,7 +83,8 @@ const DEFAULT_OPTIONS: MessageStreamOptions = {
   highWaterMark: 0,
   maxStreams: 5,
   pullTimeout: PULL_TIMEOUT,
-  timeout: 300000,
+  timeout: RESET_RETRIES_COUNT_TIME_MS,
+  maxRetries: DEFAULT_MAX_RETRIES,
 };
 
 interface StreamState {
@@ -109,21 +121,6 @@ export class StatusError extends Error implements ServiceError {
 }
 
 /**
- * Error thrown when we fail to open a channel for the message stream.
- *
- * @class
- *
- * @param {Error} err The original error.
- */
-export class ChannelError extends Error implements ServiceError {
-  code: status;
-  constructor(err: Error) {
-    super(`Failed to connect to channel. Reason: ${err.message}`);
-    this.code = err.message.includes('deadline') ? DEADLINE : UNKNOWN;
-  }
-}
-
-/**
  * @typedef {object} MessageStreamOptions
  * @property {number} [highWaterMark=0] Configures the Buffer level for all
  *     underlying streams. See
@@ -133,13 +130,16 @@ export class ChannelError extends Error implements ServiceError {
  * @property {number} [pullTimeout=900000] Timeout to be applied to each
  *     underlying stream. Essentially this just closes a `StreamingPull` request
  *     after the specified time.
- * @property {number} [timeout=300000] Timeout for establishing a connection.
+ * @property {number} [timeout=30000] Timeout for establishing a connection or
+ *     give up while retrying for max retrying in this timeout.
+ * @property {number} [maxRetries=maxStreams+3] Maximum number of times to retry.
  */
 export interface MessageStreamOptions {
   highWaterMark?: number;
   maxStreams?: number;
   pullTimeout?: number;
   timeout?: number;
+  maxRetries?: number;
 }
 
 /**
@@ -157,6 +157,9 @@ export class MessageStream extends PassThrough {
   private _options: MessageStreamOptions;
   private _streams: Map<PullStream, boolean>;
   private _subscriber: Subscriber;
+  private _retries: number;
+  private _timeSinceFirstError: number;
+
   constructor(sub: Subscriber, options = {} as MessageStreamOptions) {
     options = Object.assign({}, DEFAULT_OPTIONS, options);
 
@@ -166,13 +169,11 @@ export class MessageStream extends PassThrough {
     this._options = options;
     this._streams = new Map();
     this._subscriber = sub;
-
+    this._retries = 0;
+    this._timeSinceFirstError = Date.now();
     this._fillStreamPool();
-
-    this._keepAliveHandle = setInterval(
-      () => this._keepAlive(),
-      KEEP_ALIVE_INTERVAL
-    );
+    this._keepAliveHandle =
+        setInterval(() => this._keepAlive(), KEEP_ALIVE_INTERVAL);
     this._keepAliveHandle.unref();
   }
   /**
@@ -258,12 +259,6 @@ export class MessageStream extends PassThrough {
       this._addStream(stream);
       stream.write(request);
     }
-
-    try {
-      await this._waitForClientReady(client);
-    } catch (e) {
-      this.destroy(e);
-    }
   }
   /**
    * It is critical that we keep as few `PullResponse` objects in memory as
@@ -301,8 +296,12 @@ export class MessageStream extends PassThrough {
    */
   private _onEnd(stream: PullStream, status: StatusObject): void {
     this._removeStream(stream);
-
-    if (RETRY_CODES.includes(status.code)) {
+    if (Date.now() - this._timeSinceFirstError > this._options.timeout!) {
+      this._retries = 0;
+      this._timeSinceFirstError = Date.now();
+    }
+    if (RETRY_CODES.includes(status.code) &&
+        this._retries++ < this._options.maxRetries!) {
       this._fillStreamPool();
     } else if (!this._streams.size) {
       this.destroy(new StatusError(status));
@@ -378,22 +377,5 @@ export class MessageStream extends PassThrough {
    */
   private _setHighWaterMark(stream: PullStream): void {
     stream._readableState.highWaterMark = this._options.highWaterMark!;
-  }
-  /**
-   * Promisified version of gRPCs Client#waitForReady function.
-   *
-   * @private
-   *
-   * @param {object} client The gRPC client to wait for.
-   * @returns {Promise}
-   */
-  private async _waitForClientReady(client: ClientStub): Promise<void> {
-    const deadline = Date.now() + this._options.timeout!;
-
-    try {
-      await promisify(client.waitForReady).call(client, deadline);
-    } catch (e) {
-      throw new ChannelError(e);
-    }
   }
 }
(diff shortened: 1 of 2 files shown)


src/message-stream.ts
```diff
@@ -43,6 +43,16 @@ const DEADLINE: status = 4;
...
 
+/*!
+ * Max number of retries
+ */
+const DEFAULT_MAX_RETRIES = 3;
+
+/*!
+ * Time to reset retries counter.
+ */
+const RESET_RETRIES_COUNT_TIME_MS = 30000;
```
callmehiphop: Where does 30 seconds come from? We might want to get some input from @anguillanneuf here, but for some reason I seem to remember the PubSub team requesting that we retry for up to 5 minutes.

src/message-stream.ts
```diff
@@ -301,8 +298,14 @@ export class MessageStream extends PassThrough {
    */
   private _onEnd(stream: PullStream, status: StatusObject): void {
     this._removeStream(stream);
-
-    if (RETRY_CODES.includes(status.code)) {
+    if (Date.now() - this._timeSinceFirstError > this._options.timeout!) {
```
callmehiphop: So to clarify, this will wait to see if the stream was open for more than 30 seconds (or whatever it was set to) and if so we consider it a success?

src/message-stream.ts
ajaaym: @callmehiphop Idea is that if stream starts failing, almost all active/creating new stream will fail with in that timeout(30s by default) and it will retry for max number of retries specified(3 by default). 10-30s is good enough to determine if there is a connection/authentication issue to start with or if it needs refresh token/deadline exceeded after initial setup. 5 min to keep retrying is too long for this.

src/message-stream.ts
ajaaym: Its actually checks if there are any errors in stream try creating new stream for 30 sec or max retries and fail afterwards.

src/message-stream.ts
```diff
@@ -301,8 +298,14 @@ export class MessageStream extends PassThrough {
...
+      this._retries = 0;
+      this._timeSinceFirstError = Date.now();
+    }
+    if (
+      RETRY_CODES.includes(status.code) &&
+      this._retries++ < this._options.maxRetries!
```
callmehiphop: It isn't really uncommon for all the streams to end around the same time, I think if this were to happen then we would likely blow past the default `maxRetries` pretty quick and end up having too few streams open at any given time.

ajaaym: _fillStreamPool actually fills up stream pool, so it will create all the stream it requires to fill up the pool back even on last retry. The scenario you mentioned can happen only if there is a single stream thats connected and all other old and newly created stream keeps failing which sounds less likely. In this case it will recreate all remaining stream if last non failing stream fails.

callmehiphop: I don't think that is always true. I've seen scenarios where the ending of streams can be staggered ever so slightly and in some cases `_fillStreamPool` will need to be called multiple times.

ajaaym: chance of that scenario could be reduced by decreasing timeout and increasing maxRetries. WDYT of adding call to  _fillStreamPool  in keepAliveHandler if there are not stream failure since specified timeout.

callmehiphop: I like the idea of improving our refill strategy, but I think putting it in the keep alive handler is a little heavy handed. I think even just wrapping `fillStreamPool` in something like `setImmediate` might do the trick.

ajaaym: fillStreamPool is actually a async function and its getting executed in separate event loop than the one that failed stream.

callmehiphop: Sorry, the thought I was trying to convey is that we might want to slightly stagger the call to fill the pool so that more streams have a chance to close.

ajaaym: @callmehiphop Added fix to fill up pool if stream errors out after retries are exhausted and there are less active stream in a pool. Also changed logic for retry count to count multiple stream failure as single retry.

src/message-stream.ts
```diff
@@ -379,21 +403,4 @@ export class MessageStream extends PassThrough {
...
   }
-  /**
-   * Promisified version of gRPCs Client#waitForReady function.
-   *
-   * @private
-   *
-   * @param {object} client The gRPC client to wait for.
-   * @returns {Promise}
-   */
-  private async _waitForClientReady(client: ClientStub): Promise<void> {
```
callmehiphop: I know we talked about deleting this, so I was expecting it, do we also need to enable channel pooling?

ajaaym: That option is available by default. 
const pubsub = new PubSub(
  {
    'grpc_gcp.apiConfig': {
      "channelPool": {
        "maxSize": 5
        },
    }
  }
);

callmehiphop: Oh, this might be problematic. My understanding is that we're making the switch to the pure JS version of gRPC soon and it doesn't support gcp. @alexander-fenster can you confirm that?

alexander-fenster: That's right. If you need to use this `grpc_gcp` functionality you'll need to do two things:
1. pass C-code gRPC as a parameter to client constructor;
2. set the `grpc_gcp` options in Pub/Sub code since they will be removed from gax.

I'm going to do the same for Spanner. If Pub/Sub is also affected, we can postpone it's upgrade to gax v1.0.0 until this is fixed in Pub/Sub code (together with Spanner).

callmehiphop: Thanks @alexander-fenster! Is the plan to keep Spanner on the C version until gcp support finds its way to grpc-js?

alexander-fenster: Yes. Also, we will remove all grpc-gcp code from gax (actually, already removed, just not yet released as v1.0.0) since it's not a good place for it (and an extra dependency on grpc), so libraries that want to use grpc-gcp should depend on it directly.

callmehiphop: Good to know. I think we're going to want postpone the gax upgrade here unless @ajaaym can think of another fix that doesn't involve channel pooling.

ajaaym: @callmehiphop WDYT of option in getClient of pubsub to provide new client vs cached one? Then depending on maxStreams we can get create maxStreams/100 + 1 client.

callmehiphop: I think that sounds good. Out of curiosity have we tested against grpc-js to see if it is affected by this?

ajaaym: I was playing around with that yesterday and looks like its not that stable yet. Yes there is a limitation of 100 concurrent stream. Also I dont see option to set flow control limit which would be needed for load balancing ultra slow consumer. When testing with 100 streams I was randomly getting this error along with DEADLINE EXCEEDED. I think its better we hold on to upgrade until we have this option and well tested with different conditions.

```
Error [ERR_STREAM_WRITE_AFTER_END]: write after end
    at writeAfterEnd (_stream_writable.js:243:12)
    at ClientHttp2Stream.Writable.write (_stream_writable.js:291:5)
    at filterStack.sendMessage.then (/Users/ajaymovalia/work/node/testwithexpress2/node_modules/@grpc/grpc-js/build/src/call-stream.js:396:34)
Emitted 'error' event at:
    at onwriteError (_stream_writable.js:431:12)
    at onwrite (_stream_writable.js:456:5)
    at process._tickCallback (internal/process/next_tick.js:63:19)
```

src/message-stream.ts
callmehiphop: I can understand the need to modify our retry mechanisms to support unauthenticated requests, but this change makes me a little nervous. My concern is that we start seeing an increase of errors that kill the subscriber that would otherwise be retried within that 5 minute duration. I would be interested to know how other clients handle this retrying this specific rpc. 

@sduskis any thoughts here?

anguillanneuf: Adding @jadekler @jskeet for comments. I don't know if the 90 seconds found in GAPIC applies here.

jadekler: We definitely need to have a more holistic look at this topic. A couple of notes:

- I think an ideal state to be in is that the library creates no deadlines, instead opting to allow the user to specify a deadline. In some languages you can do this with contexts, but not all languages have contexts. We (client libs team, AC tools team) are investigating this topic later today.
- The list of codes to retry below seems too broad. I've noticed this in several pubsub clients. I've got an action item to chat with @sduskis about this. That said, immediate feedback on this PR is that I'm a thumbs down on this PR adding UNAUTHENTICATED to retry codes. It's a signal to the user that their config is borked and should not be retried.

PR #608, opened 2019-05-15
refactor: use core grpc

callmehiphop:
I think the recent upgrade to gax 1.x borked master, I think this should fix it.

- [x] Tests and linter pass


alexander-fenster:
Thanks @callmehiphop, the changes look good to me!

PR #611, opened 2019-05-16
fix: use typescript import/export for gapics

### LINKED ISSUE #610: Property 'PublisherClient' does not exist on type
I am following the sample code here. https://github.com/googleapis/nodejs-pubsub/blob/master/samples/topics.js

const {v1} = require('@google-cloud/pubsub');

  // Creates a publisher client
  const client = new v1.PublisherClient({
    // optional auth parameters
  });

When I migrate to typescript, I got this error:

src/pubsubclient.ts:7:23 - error TS2339: Property 'PublisherClient' does not exist on type 'typeof import("/Users/hangao/Zume/ingest/node_modules/@google-cloud/pubsub/build/src/index")'.

7 const pubsub = new v1.PublisherClient(credential);
                        ~~~~~~~~~~~~~~~
#### Environment details

  - OS: mac 
  - Node.js version: v8.9.4
  - npm version: 6.9.0
  - `@google-cloud/pubsub` version:     "@google-cloud/pubsub": "^0.28.1",

callmehiphop:
Fixes #610

- [x] Tests and linter pass


src/index.ts
```diff
@@ -74,7 +74,8 @@
  * @property {constructor} SubscriberClient
  *     Reference to {@link v1.SubscriberClient}.
  */
-module.exports.v1 = require('./v1');
+import v1 = require('./v1');
+export {v1};
```
sduskis: Is this change testable?

callmehiphop: I'm sure it is, but we've never made a point to test exports in any of the clients. Although I could see some value in testing exports that have no types (like this). @JustinBeckwith @bcoe WDYT?

JustinBeckwith: Just a simple `assert.ok(spanner.v1)` would be nice :)

callmehiphop: Probably was heavy handed of me, but I've renamed the old `index.ts` test file to be `pubsub.ts` to match the actual source code we're testing. Now the `index.ts` test file just includes the export tests. The overall diff will be annoying to read, but I broke it up into logical commits if you want to review commit by commit.

PR #616, opened 2019-05-18
docs: fix npm link for @google-cloud/pubsub

bkendall:
- [x] ~~Tests and linter pass~~ N/A
- [x] ~~Code coverage does not decrease (if any source code was changed)~~ N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 [npm history][1]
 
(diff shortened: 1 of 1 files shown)


callmehiphop:
@bcoe just an FYI because I think that part is generated?

JustinBeckwith:
I think this was an old bug in releasetool that's since been fixed.  We only add to the changelog, so it makes sense this would still be around.

PR #618, opened 2019-05-20
revert!: remove pullTimeout subscriber option

callmehiphop:
Relates to https://github.com/googleapis/nodejs-pubsub/issues/468#issuecomment-493661518

The PubSub team would prefer that we do not provide this option.

- [x] Tests and linter pass


callmehiphop:
@bcoe yeah, that is pretty much the gist of why

PR #622, opened 2019-05-23
fix!: issue in subscriptions.formatMetadata_

### LINKED ISSUE #623: Decouple `retainAckedMessages ` and `messageRetentionDuration`
There are two very similar fields in Subscription: `messageRetentionDuration` and `messageRetentionDuration`.  

Creating subscriptions currently couples those fields together:

```
    if (typeof metadata.messageRetentionDuration === 'number') {
      formatted.retainAckedMessages = true;
      (formatted as google.pubsub.v1.ISubscription).messageRetentionDuration = {
        seconds: metadata.messageRetentionDuration,
        nanos: 0,
      };
    }
```

Those fields should actually be independent.  Here's the definition from the proto:

```
  // Indicates whether to retain acknowledged messages. If true, then
  // messages are not expunged from the subscription's backlog, even if they are
  // acknowledged, until they fall out of the `message_retention_duration`
  // window. This must be true if you would like to
  // <a
  // href="https://cloud.google.com/pubsub/docs/replay-overview#seek_to_a_time">
  // Seek to a timestamp</a>.
  bool retain_acked_messages = 7;

  // How long to retain unacknowledged messages in the subscription's backlog,
  // from the moment a message is published.
  // If `retain_acked_messages` is true, then this also configures the retention
  // of acknowledged messages, and thus configures how far back in time a `Seek`
  // can be done. Defaults to 7 days. Cannot be more than 7 days or less than 10
  // minutes.
  google.protobuf.Duration message_retention_duration = 8;
```

The distinction in use subtle, but the meaningful  difference to the types of messages that these settings will affect.  
- `message_retention_duration` relates to how long *unacked* messages are kept for.  
- `retain_acked_messages` relates to whether or not *acked* messages should also remain in the queue.

`retain_acked_messages` has an extra cost to the customer, and should be used in cases where the messages need to be available for auditing or playback.  It needs to be explicitly turned or on off independently of the unacked messages settings.

sduskis:
Fixes #623.

The current code is this:

```
    if (typeof metadata.messageRetentionDuration === 'number') {
      formatted.retainAckedMessages = true;
      (formatted as google.pubsub.v1.ISubscription).messageRetentionDuration = {
        seconds: metadata.messageRetentionDuration,
        nanos: 0,
      };
    }
```

`formatted.retainAckedMessages = true;` shouldn't be there.

- [x] Tests and linter pass

### DIFF
--- a/system-test/pubsub.ts
+++ b/system-test/pubsub.ts
@@ -328,6 +328,7 @@ describe('pubsub', () => {
       const subName = generateSubName();
       const threeDaysInSeconds = 3 * 24 * 60 * 60;
       const callOptions = {
+        retainAckedMessages: true,
         messageRetentionDuration: threeDaysInSeconds,
         topic: '',
         name: '',
@@ -339,7 +340,7 @@ describe('pubsub', () => {
         sub!.getMetadata((err, metadata) => {
           assert.ifError(err);
 
-          assert.strictEqual(metadata!.retainAckedMessages, true);
+          assert.strictEqual(metadata!.retainAckedMessages, false);
           assert.strictEqual(
             Number(metadata!.messageRetentionDuration!.seconds),
             threeDaysInSeconds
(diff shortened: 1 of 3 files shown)


sduskis:
I'm going to create a corresponding issue to explain why this change is necessary.

system-test/pubsub.ts
```diff
@@ -354,6 +355,36 @@ describe('pubsub', () => {
       });
     });
 
+    it('should create a subscription without message retention', done => {
```
callmehiphop: I can't help but think this would be better as a unit test.

sduskis:
It looks like I need to update `SubscriptionMetadata` to explicitly include `retainAckedMessages`

sduskis:
@callmehiphop, would you mind raising a PR with this change + unit tests?  I'll close this one.

callmehiphop:
@sduskis sure

sduskis:
@callmehiphop, I'd appreciate it if you raised a PR with that change.  I trust your judgement about Node.js more than I trust mine.

PR #630, opened 2019-05-28
chore: cleanup resources for system test cases

### LINKED ISSUE #496: system tests leftovers

jiren:
Fixes #496 

- [x] Tests and linter pass

### DIFF
--- a/system-test/pubsub.ts
+++ b/system-test/pubsub.ts
@@ -193,10 +193,7 @@ describe('pubsub', () => {
     it('should honor the autoCreate option', done => {
       const topic = pubsub.topic(generateTopicName());
 
-      topic.get({autoCreate: true}, (err, _topic, _apiResponse) => {
-        assert.ifError(err);
-        done();
-      });
+      topic.get({autoCreate: true}, done);
     });
 
     it('should confirm if a topic exists', done => {
@@ -363,10 +360,7 @@ describe('pubsub', () => {
     it('should honor the autoCreate option', done => {
       const sub = topic.subscription(generateSubName());
 
-      sub.get({autoCreate: true}, (err, _sub, _apiResponse) => {
-        assert.ifError(err);
-        done();
-      });
+      sub.get({autoCreate: true}, done);
     });
 
     it('should confirm if a sub exists', done => {
--- a/system-test/pubsub.ts
+++ b/system-test/pubsub.ts
@@ -31,6 +31,8 @@ import {Policy} from '../src/iam';
 const pubsub = new PubSub();
 
 describe('pubsub', () => {
+  const TESTS_PREFIX = `pubsub-tests-${shortUUID()}`;
+
   const TOPIC_NAMES = [
     generateTopicName(),
     generateTopicName(),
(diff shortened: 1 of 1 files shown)


callmehiphop (commented):
This is a good start, but I think it is still possible we'll still see lingering resources in the event that an after hook fails.

In other client libraries our approach has been to record a timestamp representing the test run on each of the test resources, then in the after hook we grab a list of all the resources and delete any that are older than 1 hour. 

I think we can accomplish something similar here by using labels and the `setMetadata` methods.

jiren:
@callmehiphop I am planning to add a prefix to resources name and delete all resources by identifying prefix in after callback.

callmehiphop:
@jiren I'm not sure how we plan on using a prefix, but I think it will suffer from the same issues. If we only delete resources with the unique prefix and something goes wrong, then there's a good chance those resources will linger around forever. On the flip side if we were to delete any resources that didn't have the prefix, there's a good chance we could sabotage any parallel CI runs.

jiren:
> This is a good start, but I think it is still possible we'll still see lingering resources in the event that an after hook fails.
> 
> In other client libraries our approach has been to record a timestamp representing the test run on each of the test resources, then in the after hook we grab a list of all the resources and delete any that are older than 1 hour.
> 
> I think we can accomplish something similar here by using labels and the `setMetadata` methods.

@callmehiphop Right now there is no direct option available to set metadata(labels) to Topic, Subscription, Snapshot. So labels can be set using setMetadata only after resource created. In this case, any error can prevent setting metadata.
 
We should add an option to labels while creating Topic, Subscription, Snapshot as per protobuf [definition](https://github.com/googleapis/nodejs-pubsub/blob/884711f945698794d08a93e2e67bd9a2adc9fc08/protos/google/pubsub/v1/pubsub.proto#L136).

system-test/pubsub.ts
```diff
@@ -45,22 +59,67 @@ describe('pubsub', () => {
...
+    return generateName('topic');
   }
 
   function getTopicName(topic: Topic) {
     return topic.name.split('/').pop();
   }
 
+  function deleteTestResource(resource: Resource) {
+    // Delete resource from current test run.
+    if (resource.name.includes(PREFIX_WITH_RUNID)) {
```
callmehiphop: Seems like we don't need both the ID logic and the timestamp logic, I would vote we just do the timestamp.

callmehiphop (commented):
One little nitpick, otherwise LGTM!

system-test/pubsub.ts
jiren: Removed run id and updated it with current time value with with milliseconds.

system-test/pubsub.ts
```diff
@@ -134,7 +193,10 @@ describe('pubsub', () => {
     it('should honor the autoCreate option', done => {
       const topic = pubsub.topic(generateTopicName());
 
-      topic.get({autoCreate: true}, done);
+      topic.get({autoCreate: true}, (err, _topic, _apiResponse) => {
```
callmehiphop: Why did we need to change this?

jiren: In my view need to check for any error.

callmehiphop: If an error gets passed into `done` the test will fail, so the test is essentially the same - this one is just a little more verbose. I think including `topic` and `apiResponse` as params should/will trigger a lint error since they are unused.

jiren: Got it. Thanks for info.

callmehiphop (approved):

PR #633, opened 2019-05-28
docs(subscription): mention 100 stream limit

callmehiphop:
Relates to #550

- [x] Tests and linter pass

### DIFF
--- a/src/subscription.ts
+++ b/src/subscription.ts
@@ -134,6 +134,12 @@ export type SetSubscriptionMetadataResponse = MetadataResponse;
  * nack of such message when the processing is done. **Note:** message
  * redelivery is still possible.
  *
+ * By default each {@link PubSub} instance can handle 100 open streams, with
+ * default options this translates to 20 Subscriptions per PubSub instance.
+ * If you desire to use more than 20 Subscriptions you can either create
+ * multiple PubSub instances or lower the `options.streamingOptions.maxStreams`
+ * value.
+ *
  * @class
  *
  * @param {PubSub} pubsub PubSub object.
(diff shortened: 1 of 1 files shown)


callmehiphop:
@sduskis I'm not sure if we want to keep #550 until a real fix is made, but I'm going to let you make that call.

callmehiphop:
@ctavan if you have any thoughts here, please let me know!

callmehiphop:
> did I follow that this restriction doesn't exist in grpc-js

I haven't looked/tested myself, but I believe that @ajaaym said it does exist there as well.

sduskis:
@callmehiphop, 20 subscriptions * 5 streams = 100 streams.  User's won't be able to ack, since they won't be able to open a new stream.  I'm not sure what the limit ought to be, but it probably should be <20.

callmehiphop:
@sduskis good call, I went ahead and made an edit to mention that. Thanks!

ctavan:
@callmehiphop I was about to make the same comment as @sduskis since it can handle at most 19 subscriptions with default options but I'm happy with the amended text!

So the documentation update now LGTM!

However just out of curiosity: Does that mean that the PubSub client in its current implementation opens an additional stream for sending acks/nacks? And is that one additional stream per `Subscription` or one additional stream per `PubSub` client instance? The latter would match my observations of being able to subscribe with exactly 99 concurrent streams but not with 100.

ctavan:
Maybe just one nit: 

[…] with default options this translates to **strictly** less than 20 Subscriptions per PubSub […]

callmehiphop:
> Does that mean that the PubSub client in its current implementation opens an additional stream for sending acks/nacks?

@ctavan I believe this is the case. As I understand it, unary calls made via gRPC open a stream under the hood.

PR #636, opened 2019-06-01
feat: add .repo-metadata.json, start generating README.md

bcoe:
This adds a `.repo-metadata.json`, which is used to automate publishing docs when new package versions are released.

This PR also re-runs our automated README generation, which is based on `.repo-metadata.json` as well.

I've left a couple questions for folks inline.


samples/quickstart.js
```diff
@@ -15,7 +15,7 @@
 
 'use strict';
 
-// [START pubsub_quickstart_create_topic]
```
bcoe: Is renaming this region tag to the standard `quickstart` okay? CC: @sduskis, @JustinBeckwith, I wasn't sure how load bearing the region tag is, e.g., will this mean that we break samples on `cloud.google.com`?

JustinBeckwith: That is very likely to break samples.  Can you follow up with @jmdobry or @fhinkel on the right way to change this?

PR #658, opened 2019-06-13
fix(deps)!: use grpc-js instead of grpc extension

### LINKED ISSUE #627: Switch from grpc to @grpc/grpc-js
**Is your feature request related to a problem? Please describe.**

`@grpc/grpc-js` has multiple advantages over `grpc`. All Google libraries are now switching to `@grpc/grpc-js`. This is the only library left that I still see a dependency on `grpc`, which is use only in tests and as a source of type definitions.

 **Describe the solution you'd like**

I'd like to see the dependency dropped and the source code to either use type definitions from `google-gax` or `@grpc/grpc-js`.

 **Describe alternatives you've considered**

I don't see alternatives.

 **Additional context**

Not applicable.

callmehiphop:
Fixes #627 

- [x] Tests and linter pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,6 @@
     "extend": "^3.0.2",
     "google-auth-library": "^3.0.0",
     "google-gax": "^1.0.0",
-    "grpc": "1.21.1",
     "is-stream-ended": "^0.1.4",
     "lodash.snakecase": "^4.1.1",
     "p-defer": "^3.0.0",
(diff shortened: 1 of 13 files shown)


callmehiphop:
@sduskis I'm running some stress tests locally, but as of right now I think we're looking good. I wrote a new test specifically for messages that are 9mb and it appears to be working just fine.

PR #707, opened 2019-07-31
docs(subscription): prevent sub options from getting removed

### LINKED ISSUE #696: question / suggestion about flow control
I can't figure out from the docs, perhaps someone could clarify.

I'm trying to use flowControl to limit the number of concurrent messages that are delivered to my subscriber to 1.  It seems that I can use `flowControl` options in the creation of the subscription AND in the creation of the subscriber.  

Q1) Which do I use? (Does it matter?)  

From reading the source code I noticed the `allowExcessMessages` option, which looks useful.  But I can't see it documented [here](https://googleapis.dev/nodejs/pubsub/latest/global.html#CreateSubscriptionRequest).

Q2) Am I looking at the wrong docs!?  Or could it please be added.

There is an `ackDeadline` option, and a `timeout` option.  Whats the difference?

Q3) Can't see in the docs what time unit the `messageRetentionDuration` option is expressed in. Could this be added?

My code currently looks like this:

        const options = {
            flowControl: {
                maxMessages: 1,
                messageRetentionDuration: 10
            },
            allowExcessMessages: false,
            ackDeadline: 60
        };

        await pubsub.createSubscription(topicName, subscriptionName, options);

        const subscription = pubsub.subscription(subscriptionName, options);
        const [newSubscription] = await subscription.get({
            autoCreate: true,
            timeout: 50,
        });

        newSubscription.on(`message`, async (message) => {
            const data = JSON.parse(message.data);

It _feels_ wrong, any suggestions on a way to clean it up?

Thanks

callmehiphop:
Another case of TypeScript removing docs because of where they were declared.

Fixes #696 

- [x] Tests and linter pass

### DIFF
--- a/src/lease-manager.ts
+++ b/src/lease-manager.ts
@@ -19,6 +19,13 @@ import {freemem} from 'os';
 
 import {Message, Subscriber} from './subscriber';
 
+export interface FlowControlOptions {
+  allowExcessMessages?: boolean;
+  maxBytes?: number;
+  maxExtension?: number;
+  maxMessages?: number;
+}
+
 /**
  * @typedef {object} FlowControlOptions
  * @property {boolean} [allowExcessMessages=true] PubSub delivers messages in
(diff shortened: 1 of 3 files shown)


JustinBeckwith (approved):

JustinBeckwith:
Looks like it's hitting on a 404:

 Scanning http://localhost:5921/global.html
  [404] https://googleapis.github.io/gax-nodejs/CallSettings.html
  [200] http://localhost:5921/message-queues.js.html

callmehiphop:
@JustinBeckwith interesting, looks like that interface was removed from the documentation entirely. We should probably expect 404s all over the place

JustinBeckwith:
I think this got fixed elsewhere when it was broken.  The new url should be:
http://googleapis.github.io/gax-nodejs/interfaces/CallOptions.html

PR #716, opened 2019-08-09
feat: ordered messaging

callmehiphop:


src/topic.ts
```diff
@@ -592,13 +597,27 @@ export class Topic {
     );
   }
 
-  publish(data: Buffer, attributes?: Attributes): Promise<string>;
+  publish(data: Buffer, options?: PubsubMessage): Promise<[string]>;
   publish(data: Buffer, callback: PublishCallback): void;
   publish(
     data: Buffer,
-    attributes: Attributes,
+    options: PubsubMessage,
```
kamalaboulhosn: Why is the options parameter of type PubsubMessage? Is it supposed to be MessageOptions?

system-test/pubsub.ts
```diff
@@ -103,11 +103,11 @@ describe('pubsub', () => {
     const subscriptionStream = pubsub
       .getSubscriptionsStream()
       .on('data', deleteTestResource);
-    const snapshotStream = pubsub
-      .getSnapshotsStream()
-      .on('data', deleteTestResource);
+    // const snapshotStream = pubsub
```
kamalaboulhosn: Why commented out? Should the lines be removed?

src/publisher/index.ts
```diff
@@ -0,0 +1,164 @@
...
+   *
+   * @throws {TypeError} If data is not a Buffer object.
+   * @throws {TypeError} If any value in `attributes` object is not a string.
+   *
+   * @param {buffer} data The message data. This must come in the form of a
+   *     Buffer object.
+   * @param {MessageOptions} [options] Options for this message.
+   * @param {PublishCallback} [callback] Callback function.
+   */
+  publish(message: PubsubMessage, callback: PublishCallback): void {
```
kamalaboulhosn: The parameters don't seem to match the description in the comment.

src/publisher/message-queues.ts
```diff
@@ -0,0 +1,265 @@
...
+
+    if (this.inFlight || this.pending) {
+      return;
+    }
+
+    this.beginNextPublish();
+  }
+  /**
+   * Prepares the next batch to be published. If the batch is full we will
+   * publish immediately, otherwise we'll wait until the `maxMilliseconds`
```
kamalaboulhosn: If there is a publish outstanding, then we should continue to batch until that publish has completed, even if it exceeds the batch time specified in the batchOptions.

src/topic.ts
```diff
@@ -86,12 +87,16 @@ export type GetTopicSubscriptionsResponse = PagedResponse<
...
+ * @param {PublishOptions} [options] Publisher configuration object.
  *
  * @example
  * const {PubSub} = require('@google-cloud/pubsub');
  * const pubsub = new PubSub();
  *
  * const topic = pubsub.topic('my-topic');
+ *
+ * @example <caption>To enable message ordering, set `messageOrdering` to true.</caption>
+ * const topic = pubsub.topic('ordered-topic', {messageOrdering: true});
```
kamalaboulhosn: For consistency, let's call this enableMessageOrdering.

kamalaboulhosn:
This is just the changes for the publish side, correct?

callmehiphop:
@kamalaboulhosn yep, there actually isn't anything for us to do on the Subscriber side of things.

src/topic.ts
```diff
@@ -592,13 +597,27 @@ export class Topic {
     );
   }
 
-  publish(data: Buffer, attributes?: Attributes): Promise<string>;
+  publish(data: Buffer, options?: PubsubMessage): Promise<[string]>;
```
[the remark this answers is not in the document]
callmehiphop: I think in this case we should really consider it, especially since we aren't GA yet.

src/topic.ts
callmehiphop: `PubsubMessage` is just the name of the generated type we get from the protos. I can update it though since `MessageOptions` definitely sounds concise.

system-test/pubsub.ts
callmehiphop: Whoops, I needed to add a note here. List snapshots doesn't work with the emulator so I commented it out as a temporary workaround.

src/publisher/index.ts
callmehiphop: Ah! Nice find

src/publisher/message-queues.ts
callmehiphop: I believe this does that. While waiting for a publish to finish we continue to build batches based on the `maxMessages` and `maxBytes` options (maybe that is wrong?) and essentially build a queue of batches. Once a publish is complete we check to see if the next batch is either full or if we hit the `maxMilliseconds`.

Thinking about it now, while a pending publish is in place we should continue to batch until the API limit is met, which in turn means we ignore the user limits in this scenario. Does that sound right?

src/topic.ts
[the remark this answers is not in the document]
kamalaboulhosn: If we are going to change the API, then I recommend we get rid of the data field altogether and just have someone send in a PubsubMessage object that we can pass through to publish itself. That way, anything else we add to the message won't require an API change.

bcoe:
@callmehiphop this is looking reasonable to me, mind enumerating the breaking changes in the PR body; is it mostly TypeScript definitions that change?

src/topic.ts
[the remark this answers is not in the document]
callmehiphop: @kamalaboulhosn re: breaking change

So it would actually pretty easy to avoid a breaking change here with the following overloads

```ts
publish(data: Buffer, attributes?: object)
publish(message: PubsubMessage)
```

The reason I was leaning at breaking at first was because we have a second publish method called [`publishJSON`](https://googleapis.dev/nodejs/pubsub/latest/Topic.html#publishJSON) that already accepts an object as the first param, so it starts to get a little murky there.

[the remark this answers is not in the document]
kamalaboulhosn: Is there a typical way that one would do this in node.js? In Python, we could add an optional argument for the ordering key and that would be backwards compatible. If that's an option here and fits with convention, we may want to consider it. Otherwise, the new method that just takes the PubsubMessage seems reasonable to me.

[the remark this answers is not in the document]
callmehiphop: I think generally we would either overload or use an object in a similar way to what I'm proposing.

I'm also fine with adding a new method with a deprecation notice on `publish`/`publishJSON`. Any opinions on a name?

@bcoe @stephenplusplus do either of you have any feelings here?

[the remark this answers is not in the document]
kamalaboulhosn: In that case, I'm fine going with the new method that just takes the PubsubMessage as you have proposed. We can have the different versions for now and then at some future point deprecate the ones we don't want people to use. I don't think supporting the multiple versions will be too much of a burden for us since they can all ultimately call the one that takes the PubsubMessage.

src/topic.ts
```diff
@@ -717,8 +717,100 @@ export class Topic {
...
+   * };
+   *
+   * topic.publishMessage({data}, callback);
+   *
+   * @example <caption>Publish JSON message data.</caption>
+   * const json = {foo: 'bar'};
+   *
+   * topic.publishMessage({json}, callback);
+   *
+   * @example <caption>To publish messages in order, make sure message ordering is enabled and provide an ordering key</caption>
```
kamalaboulhosn: Can we also mention that this feature is currently experimental? We use this language in the Java library: "Ordering is not yet fully supported and requires special project enablements."

callmehiphop:
@stephenplusplus @bcoe @kamalaboulhosn just wrapped testing all the things here with some light refactoring. If you want to take another look through that would be excellent, because I think we're ready to go!

callmehiphop:
@bcoe @JustinBeckwith CI is getting shut down by the Snapshot error. Anything we can do to merge this? I'd like to cut a release soon, preferably earlier in the week to not mess up any weekend warriors.

PR #717, opened 2019-08-12
use GOOGLE_CLOUD_PROJECT env var with emulator

### LINKED ISSUE #582: feat: do not require projectId to be set in emulator
It can be [confusing](https://github.com/googleapis/nodejs-pubsub/issues/571) for folks that `projectId` needs to be explicitly set, given that `project_id` is populated in generated credentials for a service.

It would be nice if when running in the emulator behavior was the same.

derhuerst:
Fixes #582

- [x] Tests and linter pass

### DIFF
--- a/src/pubsub.ts
+++ b/src/pubsub.ts
@@ -276,7 +276,12 @@ export class PubSub {
     this.determineBaseUrl_();
     this.api = {};
     this.auth = new GoogleAuth(this.options);
+
     this.projectId = this.options.projectId || PROJECT_ID_PLACEHOLDER;
+    if (!('projectId' in this.options) && process.env.PUBSUB_EMULATOR_HOST && process.env.GOOGLE_CLOUD_PROJECT) {
+      this.options.projectId = process.env.GOOGLE_CLOUD_PROJECT;
+    }
+
     if (this.options.promise) {
       this.Promise = this.options.promise;
     }
--- a/test/pubsub.ts
+++ b/test/pubsub.ts
@@ -282,6 +282,23 @@ describe('PubSub', () => {
     it('should localize a Promise override', () => {
       assert.strictEqual(pubsub.Promise, OPTIONS.promise);
     });
+
+    describe('with PUBSUB_EMULATOR_HOST & GOOGLE_CLOUD_PROJECT environment vars', () => {
+      before(() => {
+        process.env.PUBSUB_EMULATOR_HOST = 'localhost:9090';
+        process.env.GOOGLE_CLOUD_PROJECT = 'some-google-cloud-project';
+      });
+
+      after(() => {
+        delete process.env.PUBSUB_EMULATOR_HOST;
+        delete process.env.GOOGLE_CLOUD_PROJECT;
+      });
+
+      it('should use the GOOGLE_CLOUD_PROJECT env var', () => {
+        const pubsub = new PubSub();
+        assert.strictEqual(pubsub.options.projectId, 'some-google-cloud-project');
+      });
+    });
   });
 
   describe('createSubscription', () => {


derhuerst:
@googlebot I signed it!

stephenplusplus:
Ahh, I think I know the issue here. We do support that env var, but we do so from the authentication library, [google-auth-library](http://gitnpm.com/google-auth-library). We should probably do something like this in our base class that all of our libraries inherit from, which is the `Service` class in nodejs-common. [Here's the line](https://github.com/googleapis/nodejs-common/blob/2dd8f5c7bbef76901eb15b7cd8c56b3a8ce22f57/src/service.ts#L112) that could probably just be:

```
this.projectId = options.projectId || process.env.GOOGLE_CLOUD_PROJECT || PROJECT_ID_TOKEN;
```

Would you mind PR'ing over there? Sorry for the trouble.

derhuerst:
> We should probably do something like this in our base class that all of our libraries inherit from, which is the Service class in nodejs-common.

Even then [this library would still overwrite `projectId`](https://github.com/googleapis/nodejs-pubsub/blob/b9408118dbc5e4d8c580e23a586c0e19caeae0db/src/pubsub.ts#L279) AFAICT.

> Would you mind PR'ing over there? Sorry for the trouble.

I do actually. I don't have the capacity for it right now.

stephenplusplus:
I just realized this library doesn't use common at all, so let me go ahead and pretend I didn't say anything. I'll let others closer to this codebase hop in to review this.

src/pubsub.ts
```diff
@@ -276,7 +276,12 @@ export class PubSub {
     this.determineBaseUrl_();
     this.api = {};
     this.auth = new GoogleAuth(this.options);
+
     this.projectId = this.options.projectId || PROJECT_ID_PLACEHOLDER;
+    if (!('projectId' in this.options) && process.env.PUBSUB_EMULATOR_HOST && process.env.GOOGLE_CLOUD_PROJECT) {
```
callmehiphop: Could you expand on this a little bit? Seems like this PR is hoping to explicitly prefer the `GOOGLE_CLOUD_PROJECT` env var only when using the emulator. I think this might end up giving unexpected behavior when not using the emulator. 

Why not just do something like

```js
this.projectId = this.options.projectId || process.env.GOOGLE_CLOUD_PROJECT || PROJECT_ID_PLACEHOLDER;
```

test/pubsub.ts
```diff
@@ -282,6 +282,23 @@ describe('PubSub', () => {
     it('should localize a Promise override', () => {
       assert.strictEqual(pubsub.Promise, OPTIONS.promise);
     });
+
+    describe('with PUBSUB_EMULATOR_HOST & GOOGLE_CLOUD_PROJECT environment vars', () => {
+      before(() => {
+        process.env.PUBSUB_EMULATOR_HOST = 'localhost:9090';
```
callmehiphop: Assuming these env vars are set when some one runs the test these hooks would remove them entirely. It would be good to cache the current values and in the after hook reset them to said values.

src/pubsub.ts
derhuerst: It doesn't prefer it. It only uses `GOOGLE_CLOUD_PROJECT` for `projectId` if it hasn't been passed manually.

> Why not just do something like
>
> ```js
> this.projectId = this.options.projectId || process.env.GOOGLE_CLOUD_PROJECT || PROJECT_ID_PLACEHOLDER;
> ```

Because this would override manually passed *falsy values*, such as `''`, `0`, `null`, `undefined`. It didn't want to make the assumption of interpreting falsy values as "pick the env var or the default one".

test/pubsub.ts
derhuerst: I copied this code from below. I'm fine with adding it.

In general I would make the assumption that each test (or its registered `before`/`beforeAll`) sets up the whole environment. Otherwise your tests rely on a specific order of execution.

Also keep in mind that (re-)setting them to `undefined` is not the same as just calling `delete`.

src/pubsub.ts
callmehiphop: If the projectId is not set we attempt to pull it in from the auth library. Setting it in this way (when using the emulator) means that we are preferring it over whatever the auth library would otherwise provide.

callmehiphop: Ah, my mistake, looks like we don't pull the project id from the auth library when using the emulator. I think maybe a better fix would be to always pull the project ID from the auth library regardless of whether or not we're using the emulator. 

@stephenplusplus @bcoe thoughts?

test/pubsub.ts
derhuerst: What should I do here? Add the reset?

src/pubsub.ts
VinceOPS: Hi there, I might be wrong but I assume that the variable should be `PUBSUB_PROJECT_ID`? Just like there is `DATASTORE_PROJECT_ID` when using the Datastore emulator?

Thanks!

callmehiphop:
I opened #731 as an alternative fix based on some of the conversation that came up in the code review. If anyone wants to test the fix, that would be great!

PR #731, opened 2019-09-04
fix: pull projectId from auth client with emulator

### LINKED ISSUE #582: feat: do not require projectId to be set in emulator

callmehiphop:
Fixes #582
Closes #717 

- [x] Tests and linter pass

### DIFF
--- a/src/pubsub.ts
+++ b/src/pubsub.ts
@@ -889,7 +889,7 @@ export class PubSub {
   getClient_(config: GetClientConfig, callback: GetClientCallback) {
     const hasProjectId =
       this.projectId && this.projectId !== PROJECT_ID_PLACEHOLDER;
-    if (!hasProjectId && !this.isEmulator) {
+    if (!hasProjectId) {
       this.auth.getProjectId((err, projectId) => {
         if (err) {
           callback(err);
--- a/src/pubsub.ts
+++ b/src/pubsub.ts
@@ -526,6 +526,7 @@ export class PubSub {
     if (!apiEndpoint && !process.env.PUBSUB_EMULATOR_HOST) {
       return;
     }
+
     const baseUrl = apiEndpoint || process.env.PUBSUB_EMULATOR_HOST;
     const leadingProtocol = new RegExp('^https*://');
     const trailingSlashes = new RegExp('/*$');
@@ -537,6 +538,10 @@ export class PubSub {
     this.options.port = baseUrlParts[1];
     this.options.sslCreds = grpc.credentials.createInsecure();
     this.isEmulator = true;
+
+    if (!this.options.projectId && process.env.PUBSUB_PROJECT_ID) {
+      this.options.projectId = process.env.PUBSUB_PROJECT_ID;
+    }
   }
 
   getSnapshots(options?: PageOptions): Promise<GetSnapshotsResponse>;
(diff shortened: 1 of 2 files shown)


bcoe (approved):
LGTM, assuming this is tested 👍

hrgui:
## Discovered test failure

I tested this change with only `process.env.PUBSUB_EMULATOR_HOST = "localhost:8538";`
`const pubsub = new PubSub();`, and it gave me the following error:

```
(node:49695) UnhandledPromiseRejectionWarning: Error: Unable to detect a Project Id in the current environment. 
To learn more about authentication and Google APIs, visit: 
https://cloud.google.com/docs/authentication/getting-started
    at _getDefaultProjectIdPromise.Promise (/projects/pubsub-emulator-test/node_modules/google-auth-library/build/src/auth/googleauth.js:90:31)
    at process._tickCallback (internal/process/next_tick.js:68:7)
(node:49695) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:49695) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

The environment variable `GOOGLE_CLOUD_PROJECT` is not set. When setting the environment variable `GOOGLE_CLOUD_PROJECT to `moot``, the change works.

## Reason why this test is valid
As per https://cloud.google.com/pubsub/docs/emulator#manually_setting_the_variables, it says the following:

> you only need to set PUBSUB_EMULATOR_HOST for your application to connect to the emulator


## Repo

1. Run emulator on port 8538
2. `node publisher.js`
```js
process.env.PUBSUB_EMULATOR_HOST = "localhost:8538";
// process.env.GOOGLE_CLOUD_PROJECT = "moot"; uncomment this and it works

const {PubSub} = require('@google-cloud/pubsub');

const pubsub = new PubSub();

async function init() {
  try {
    await pubsub.createTopic('test');
  } catch(e) {

  }

  await pubsub.topic('test').publishJSON({
    date: Date.now()
  });
}

init();

```
3. `node subscriber.js`:
```js
process.env.PUBSUB_EMULATOR_HOST = "localhost:8538";
// process.env.GOOGLE_CLOUD_PROJECT = "moot"; uncomment this and it works

const {PubSub} = require('@google-cloud/pubsub');


const pubsub = new PubSub();

async function init() {
  try {
    await pubsub.topic('test').createSubscription('test');
  } catch(e) {
    // topic already created or subscription already created
    console.error(e);
  }

  const subscription = pubsub.subscription('test');
  subscription.on('message', (msg) => {
    msg.ack();
    console.log(JSON.parse(msg.data));
  });

  subscription.on('error', (e) => {
    console.error(e);
  });
}

init();
```

src/pubsub.ts
```diff
@@ -889,7 +889,7 @@ export class PubSub {
   getClient_(config: GetClientConfig, callback: GetClientCallback) {
     const hasProjectId =
       this.projectId && this.projectId !== PROJECT_ID_PLACEHOLDER;
-    if (!hasProjectId && !this.isEmulator) {
+    if (!hasProjectId) {
```
hrgui: I wonder if the check `!this.isEmulator` should still be left in, because it is valid that this should pull from auth client when it is not in emulator.

Now when it is emulator mode, is it valid to pull from auth? Right below line 903, i did something like so:
```js
        if (!hasProjectId && this.isEmulator) {
            const {GOOGLE_CLOUD_PROJECT, PUBSUB_PROJECT_ID} = process.env;
            const projectId = PUBSUB_PROJECT_ID || GOOGLE_CLOUD_PROJECT || 'pubsub-emulator';
            this.projectId = projectId;
            this.options.projectId = projectId;
        }
```

and it seems to work out fine (user can set PUBSUB_PROJECT_ID || GOOGLE_CLOUD_PROJECT or neither, because it falls back to a project id called 'pubsub-emulator' that I made up to ensure that the subscriber and the publisher are in the same project.)

However, this raises the question for the documentation from https://cloud.google.com/pubsub/docs/emulator#automatically_setting_the_variables:

> You can optionally set the PUBSUB_PROJECT_ID environment variable for the project you want to use for the emulator, but you only need to set PUBSUB_EMULATOR_HOST for your application to connect to the emulator

In my proposal, if I changed

```js
const projectId = PUBSUB_PROJECT_ID || GOOGLE_CLOUD_PROJECT || 'pubsub-emulator';
```

to 

```js
const projectId = PUBSUB_PROJECT_ID || GOOGLE_CLOUD_PROJECT;
```

the topic and subscription can't be found despite it being created, which I didn't understand - so there's a discrepancy between this package and the documentation itself.

callmehiphop:
@hrgui thanks for testing this! I've made a couple of commits based on your results and I was curious if you'd be willing to test them out?

src/pubsub.ts
callmehiphop: We actually went down this route as a feature request from users wanting to use the project ID returned from the auth client. I've made a commit that still requests the project ID from the auth client but ignores any errors in the event that the user is using the emulator.

hrgui:
@callmehiphop is the way to test it properly to pull the changes down and run `npm compile` and do `npm link`?

The changes before were simple enough for me to just change the dist files directly

hrgui:
**Tests conducted**

1. no project_id => ok
2. project_id as GOOGLE_CLOUD_PROJECT => ok (tested both sub and pub with different projects => showed expected behavior)
3. project_id as PUBSUB_PROJECT_ID => ok (tested both sub and pub with different projects => showed expected behavior)

Seems good.

callmehiphop:
@hrgui awesome, thanks for taking a look!

@bcoe I refactored some stuff since you last looked, mind taking another crack at it?

bcoe (approved):

PR #745, opened 2019-09-16
chore!: set release level to GA

callmehiphop:
Not sure if this is the right commit message or how release-please knows to make minor bumps, but hopefully this does the trick!

### DIFF
--- a/.repo-metadata.json
+++ b/.repo-metadata.json
@@ -4,7 +4,7 @@
   "product_documentation": "https://cloud.google.com/pubsub/docs/",
   "client_documentation": "https://googleapis.dev/nodejs/pubsub/latest",
   "issue_tracker": "https://issuetracker.google.com/savedsearches/559741",
-  "release_level": "beta",
+  "release_level": "ga",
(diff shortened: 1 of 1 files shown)


bcoe (approved):

alexander-fenster:
🎉 🎈

alexander-fenster:
I would vote for `chore!:` in the title, there is no refactor here :)

JustinBeckwith (commented):
So I'm not saying we _shouldn't_ do this, but I want to ask - how do you feel about the risk of breaking changes or stability with the ordered messaging PR?  That PR looks a little spicy, and I just want to make 100% sure we're not going to go GA, then find out we need to change the API again.

callmehiphop:
@JustinBeckwith we've talked about that subject in the PubSub weekly meeting for a little while now and we're all in agreement that a surface change should not be necessary and I'm committed to that.

kbariotis:
Hi everyone, I just wanted to ask, how is this a breaking change?

PR #778, opened 2019-10-09
build: adding benchmarking script

### LINKED ISSUE #776: StreamingPull Retries on OK
https://github.com/googleapis/nodejs-pubsub/blob/b9122763d85d08335a86559e8299c3cebfe5372a/src/message-stream.ts#L60-L71

[The gRPC "OK" code](https://github.com/grpc/grpc/blob/master/doc/statuscodes.md) indicates success and should not be retried.

hongalex:
This PR adds a gRPC server which wraps the pubsub client library, specifically testing streaming pull. A load testing framework can hit the gRPC server defined here to test the latency of our client library, which simply acks messages as they come in.
 
This requires the removal of grpc.OK as a retryable error, which is not currently used per https://github.com/googleapis/nodejs-pubsub/issues/776#issuecomment-540210217.

For reference: benchmark wrapper for the storage lib: https://github.com/googleapis/nodejs-storage/pull/788.

Fixes #776  🦕

### DIFF
--- a/bin/benchwrapper.js
+++ b/bin/benchwrapper.js
@@ -1,0 +1,54 @@
+const grpc = require('grpc');
+const protoLoader = require('@grpc/proto-loader');
+const {PubSub} = require('../build/src');
+
+const argv = require('yargs')
+  .option('port', {
+    description: 'The port that the Node.js benchwrapper should run on.',
+    type: 'number',
+    demand: true,
+  })
+  .parse();
+
+const PROTO_PATH = __dirname + '/pubsub.proto';
+// Suggested options for similarity to existing grpc.load behavior.
+const packageDefinition = protoLoader.loadSync(PROTO_PATH, {
+  keepCase: true,
+  longs: String,
+  enums: String,
+  defaults: true,
+  oneofs: true,
+});
+const protoDescriptor = grpc.loadPackageDefinition(packageDefinition);
+const pubsubBenchWrapper = protoDescriptor.pubsub_bench;
+
+const client = new PubSub();
+
+function recv(call, callback) {
+  const subName = call.request.sub_name;
+
+  const subOptions = {
+    streamingPull: {
+      maxStreams: 1,
+    },
+  };
+
+  const sub = client.subscription(subName, subOptions);
+
+  sub.on('message', message => {
+    message.ack();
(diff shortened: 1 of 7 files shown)


bin/benchwrapper.js
```diff
@@ -0,0 +1,70 @@
...
+const pubsubBenchWrapper = protoDescriptor.pubsub_bench;
+
+const client = new PubSub();
+
+function recv(call, callback) {
+  const subName = call.request.sub_name;
+
+  const subOptions = {
+    streamingPull: {
+      maxStreams: 1,
```
callmehiphop: I think you might want to change this to 

```js
const sub = client.subscription(subName);

sub.setOptions({
  streamingOptions: {
    maxStreams: 1
  }
});
```

bin/benchwrapper.js
```diff
@@ -0,0 +1,70 @@
...
+  };
+
+  const sub = client.subscription(subName, subOptions);
+
+  sub.on('message', message => {
+    message.ack();
+  });
+
+  sub.on('error', err => {
+    callback(null, null);
```
callmehiphop: Do we need to execute the callback after some kind of happy path too?

bin/benchwrapper.js
```diff
@@ -0,0 +1,70 @@
...
+  });
+}
+
+const server = new grpc.Server();
+
+server.addService(pubsubBenchWrapper['PubsubBenchWrapper']['service'], {
+  Recv: recv,
+});
+console.log('starting on localhost:' + argv.port);
```
callmehiphop: Totally a nitpick, so feel free to ignore/resolve this, buuut if you're feeling fancy we have template strings now!

```js
`localhost: ${argv.port}`
```

bin/benchwrapper.js
hongalex: This actually is the happy path. Closing of the stream is done by the server, which always throws an error, even if it's a success code like grpc's `OK`.

callmehiphop: Ah, right! Would you mind adding a comment that mentions that? I could see this throwing some one off in the future

hongalex: Done!

Issue #821, opened 2019-11-19
the first call to snapshot.create() fails with 13 INTERNAL


bcoe:
## The Issue

The first call to `subscription.createSnapshot` or `await snapshot.create` consistently fails with:

```
13 INTERNAL: A service error has occurred.
```

This error is reproducible in the console:

![example](https://user-images.githubusercontent.com/194609/69109625-2fcd2100-0a2d-11ea-9fe8-76942215f8f5.png)

## Workaround

Calling `create` a second time works both from the client library and console.

Issue #826, opened 2019-11-24
system and samples tests failing


alexander-fenster:
This Saturday some system and samples tests started to time out. I'm not sure if it's caused by a service issues or this library, but the problems reproduce in nightly tests, on my local machine, and in gax pubsub tests.

Example of the failure: https://source.cloud.google.com/results/invocations/55727ac5-4208-49e2-a5bd-ee92659740ca/targets/cloud-devrel%2Fclient-libraries%2Fnodejs%2Fnightly%2Fgoogleapis%2Fnodejs-pubsub%2Fnode10%2Fsystem-test/log

(the list of failing tests is consistent between runs)

stephenplusplus:
The same tests that fail in that log aren't the same for me locally. I do have one failing test:

```
  1) pubsub
       Snapshot
         seeking
           should seek to a snapshot:
     Uncaught AssertionError [ERR_ASSERTION]: ifError got unwanted exception: 6 ALREADY_EXISTS: Resource already exists in the project (resource=gcloud-tests-snapshot-6dc2ee71-1574603452115).
      at subscription.createSnapshot (system-test/pubsub.ts:730:18)
      at Immediate.request [as _onImmediate] (src/subscription.ts:464:11)
      at Object.callErrorFromStatus (node_modules/@grpc/grpc-js/build/src/call.js:30:26)
      at Http2CallStream.call.on (node_modules/@grpc/grpc-js/build/src/client.js:96:33)
      at process.nextTick (node_modules/@grpc/grpc-js/build/src/call-stream.js:75:22)
      at process.internalTickCallback (internal/process/next_tick.js:70:11)
```

PR sent to fix that: #827

bcoe:
this was an internal API issue that should now be fixed.

PR #827, opened 2019-11-24
test(system): fix snapshot tests

stephenplusplus:
RE: #826 (not sure it's a fix yet)
Fixes: #821 

This test was failing locally for me:

```
  1) pubsub
       Snapshot
         seeking
           should seek to a snapshot:
     Uncaught AssertionError [ERR_ASSERTION]: ifError got unwanted exception: 6 ALREADY_EXISTS: Resource already exists in the project (resource=gcloud-tests-snapshot-6dc2ee71-1574603452115).
      at subscription.createSnapshot (system-test/pubsub.ts:730:18)
      at Immediate.request [as _onImmediate] (src/subscription.ts:464:11)
      at Object.callErrorFromStatus (node_modules/@grpc/grpc-js/build/src/call.js:30:26)
      at Http2CallStream.call.on (node_modules/@grpc/grpc-js/build/src/client.js:96:33)
      at process.nextTick (node_modules/@grpc/grpc-js/build/src/call-stream.js:75:22)
      at process.internalTickCallback (internal/process/next_tick.js:70:11)
```

This throws back to changes made in #815 and #821. I changed the logic to work more like I think we were expecting.

The tests in the PR continue to fail as noted in #826. Unfortunately, I'm not sure how to address that yet.

### DIFF
--- a/samples/system-test/subscriptions.test.js
+++ b/samples/system-test/subscriptions.test.js
@@ -27,7 +27,8 @@ const execPromise = cmd =>
     cp.exec(cmd, {encoding: 'utf-8'}, (err, stdout, stderr) => {
       if (err) {
         err.stderr = stderr;
-        return reject(err);
+        reject(err);
+        return;
       }
       resolve(stdout);
     });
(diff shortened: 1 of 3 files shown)


callmehiphop:
@stephenplusplus any reason why we're moving away from async/await in favor of then blocks?

Issue #733, opened 2019-09-05
Samples style refresh


anguillanneuf:
- Please split large files that contain more than one sample into small files that contain a single sample per file.

- Please make code that's enclosed in a pair of region tags runnable as is by following the updated style guide. 

```node
function main(arg1 = 'default_value_one', arg2 = 1234) {
  // [START your_region_tag]
  /**
   * TODO(developer): Uncomment these variables before running the sample.
   */
  // const arg1 = 'default_value_one';
  // const arg2 = 1234;

  // Imports the Google Cloud Some API library
  const {SomeApiClient} = require('@google-cloud/some-api');

  // Instantiates a client
  const someApiClient = new SomeApiClient();

  async function doSomething() {
    // Construct request
    const request = {
      value1: arg1,
      value2: arg2,
      value3: 'hard-coded value',
    };

    const [response] = await someApiClient.doSomething(request);

    ...
  }

  doSomething();
  // [END your_region_tag]
}

main(...process.argv.slice(2));
```

PR #844, opened 2020-01-03
refactor(samples): move into separate files, and update to the latest sample style standard

feywind:
Fixes https://github.com/googleapis/nodejs-pubsub/issues/733

Refactor the topics and subscriptions samples so that they aren't single files, and so they have the new GCP sample style. I've worked to keep the same command line format for the original sample runners so we don't have to update docs immediately or confuse existing users. This involved a new test common piece to paper over the differences between running from the big sample runner vs as an individual sample.

There's still one TODO in here, which is that one of the samples was missing document tags. I think that needs to be added to the docs for full coverage. I think it goes here:

https://cloud.google.com/pubsub/docs/admin


feywind:
This is still marked as draft because I need to update the docs to include the sample that was missing doc tags, but it's otherwise pretty much done.

feywind:
It looks like the missing doc tags is a larger problem. e.g. the C# samples are also missing tags for that sample.

https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/master/pubsub/api/PubsubSample/Program.cs
(search for GetSubscription)

So I'm just going to call this ready, but I'll leave the TODO in there if that's okay. I do think we will want to fix that not-quite-a-sample, but it's going to require work across multiple repos.

samples/subscriptions/createPushSubscription.js
```diff
@@ -0,0 +1,68 @@
...
+      },
+    };
+
+    await pubSubClient
+      .topic(topicName)
+      .createSubscription(subscriptionName, options);
+    console.log(`Subscription ${subscriptionName} created.`);
+  }
+
+  createPushSubscription();
```
callmehiphop: Should we add error handling to all the samples? The old samples used top level async functions, where as with this format if an error were to occur it would result in an unhandled rejection (I think). It is probably ok, but I'm not sure if we care about that scenario or not

samples/common/common.js
```diff
@@ -0,0 +1,118 @@
...
+    this.exampleText = this.exampleText || [];
+    this.exampleText.push(exampleText);
+    return this;
+  }
+
+  // Call this last; it will either run the sample with what's been
+  // given (if it's running as main) or export this object for the
+  // consolidated runner module to use. 'forceMain' is for testing
+  // purposes.
+  execute(sampleModule, executionCallback, forceMain = false) {
```
callmehiphop: Is there a new sample requirement to allow the samples to be runnable via cli and requirable? I like it, I just haven't seen anything similar in other modules. If that is the case it might be worth bundling this up as its own module that we can import in other projects.

samples/subscriptions/createPushSubscription.js
feywind: The updated style guide isn't super clear on that point, but it does seem like something we should do. It should be super easy with the way everything is being routed through sampleMain now, so I could just throw that in.

samples/common/common.js
feywind: This is another point where I made some creative interpretation. The style guide says the samples should be in separate files so they can be run independently, but I felt like it was important to keep from pulling the rug out from under users, so I wanted to make the original topics.js and subscriptions.js keep working.

The other consideration there is just that some of the samples require numeric and/or optional arguments, and it was a lot easier to keep using yargs for that. And if I'm doing that anyway, I might as well make it plug into those original files.

callmehiphop: I see. I am definitely all for backwards compatibility, but I haven't observed that happening in other repos that have split the samples. I would be curious to hear @JustinBeckwith's take on this

feywind: Me too! It's pretty easy to just take out the aggregate sample files if we want to simplify it, but I might still be interested in keeping the common argument handling even if we do that (because of the numeric args and help text).

samples/subscriptions/createPushSubscription.js
feywind: I spent a while thinking about this and pondering the samples style guide, and I think there isn't a way to fulfill both. The style guide is pretty explicit that samples should have an async function which is immediately called at the end of the sample, but given that, there isn't a simple way to let exceptions/rejections bubble up. So I think for now I might just pass on fixing this. With Node 12 it gives a warning about unhandled rejections, but it does still print a stack trace, so that probably covers it for now...

samples/common/common.js
feywind: I re-read the style guide again and it actually says we're deprecating aggregate sample runners that have full command lines. Basically the emphasis is on making the samples easily paste-capable in the docs site. So this might be overengineered at this point. I'll bug Justin about it.

samples/common/index.js
```diff
@@ -0,0 +1,23 @@
...
+// limitations under the License.
+
+/**
+ * Package index for common sample pieces.
+ */
+
+'use strict';
+
+module.exports = {
+  ...require('./common'),
```
bcoe: stylistically, I'd probably just make the contents of `common.js`, `index.js`, when there's only a single export in a folder.

samples/subscriptions/createSubscription.js
```diff
@@ -0,0 +1,59 @@
...
+    await pubSubClient.topic(topicName).createSubscription(subscriptionName);
+    console.log(`Subscription ${subscriptionName} created.`);
+  }
+
+  createSubscription();
+  // [END pubsub_create_pull_subscription]
+}
+
+const {sampleMain} = require('../common');
+sampleMain()
```
bcoe: I think you might be able to avoid the wrapper class with something like this:

```
yargs.command('create <topicName> <subscriptionName>', 'creates a new subscription', (yargs) => {
  yargs.positional('topicName', {
    number: true // just an example of how to configure positionals
  })
}, (opts) => {
  main(opts.topicName, opts.subscriptionName);
})
.example('my-topic worker-1')
```

:point_up: I think using this shorthand for defining a `command` and `handler`, with the `.positional` feature for defining types, you could probably get away with using yargs directly.

feywind: I'm going ahead and removing the yargs usage. Only a few samples use numeric parameters, and it's simple enough to throw in a call to Number() to make sure they're the right type.

samples/common/index.js
feywind: I'm also removing common.js and its index.js, so this is going away.

samples/subscriptions/createPushSubscription.js
feywind: I changed my mind on this. It's still not 100% in line with the style guide, but it makes sense as part of a sample snippet anyway. In the call to the async "meat" function, I'm just adding a .catch(console.error) to print any errors.

README.md
```diff
@@ -104,8 +104,6 @@ has instructions for running the samples.
 | Sample                      | Source Code                       | Try it |
 | --------------------------- | --------------------------------- | ------ |
 | Quickstart | [source code](https://github.com/googleapis/nodejs-pubsub/blob/master/samples/quickstart.js) | [![Open in Cloud Shell][shell_img]](https://console.cloud.google.com/cloudshell/open?git_repo=https://github.com/googleapis/nodejs-pubsub&page=editor&open_in_editor=samples/quickstart.js,samples/README.md) |
-| Subscriptions | [source code](https://github.com/googleapis/nodejs-pubsub/blob/master/samples/subscriptions.js) | [![Open in Cloud Shell][shell_img]](https://console.cloud.google.com/cloudshell/open?git_repo=https://github.com/googleapis/nodejs-pubsub&page=editor&open_in_editor=samples/subscriptions.js,samples/README.md) |
```
JustinBeckwith: I'm surprised the list here and in `samples/README.md` went down, since those should be generated.  Can you install and try running `synthtool` to get the README generation stuff kickin?  
https://github.com/googleapis/synthtool/

samples/common/common.js
feywind: We had a talk internally and decided to go ahead and remove the aggregate sample runners, and fully simplify this all down. I've just changed the system-test modules to run the individual samples directly.

samples/subscriptions/createPushSubscription.js
```diff
@@ -0,0 +1,62 @@
...
+ * subscriptions with the Google Cloud Pub/Sub API.
+ *
+ * For more information, see the README.md under /pubsub and the documentation
+ * at https://cloud.google.com/pubsub/docs.
+ */
+
+'use strict';
+
+// Creates a new push subscription.
+function main(
```
JustinBeckwith: For each of these that take parameters - a preamble like this ensures the generated README and such are nice to use:

```js

// sample-metadata:
//   title: Export Assets
//   description: Export asserts to specified dump file path.
//   usage: node exportAssets.js <gs://my-bucket/my-assets.txt>
```

It would be nice to have one of these for each sample if possible.  For an example, see:
https://github.com/googleapis/nodejs-asset/blob/master/samples/exportAssets.js

README.md
feywind: It looks like I had the wrong version of synthtool, and it wasn't finding the samples because they were in subdirs. I just moved them up, and it's working now.

samples/subscriptions/createPushSubscription.js
feywind: This is actually an interesting case for an unrelated reason: I'm pretty sure exceptions still won't be caught when running it from the command line. The main() function is async, but it's not calling await on the inner function, so it's essentially returning a Promise that immediately resolves. It might be something we should revisit in the style guide, because I'm not sure I see a good way to catch the errors from samples unless we tack a .catch() on the call to the inner function or await on it (in which case the sample snippet is weird if pasted elsewhere).

Back to the topic at hand: this is nice, I didn't know you could add these metadata comments. I'll do that.

JustinBeckwith: It's a really good point.  This is an underspecified part of the sample style guide that we should probably look at designing out further .. FYI @fhinkel

JustinBeckwith:
This is looking great!

samples/subscriptions/createPushSubscription.js
fhinkel: If we don't await, don't we get an uncaught exception and abort?

feywind: I made an issue to track this over there:
https://github.com/googleapis/nodejs-asset/issues/242

I'm going to resolve this conversation over here since the metadata bits got updated.

Issue #855, opened 2020-01-17
Synchronous pull sample is actually synchronous pull with lease management


feywind:
This was discovered by way of a documentation bug report, which was fixed by (https://github.com/googleapis/nodejs-pubsub/pull/844). The sample called synchronousPull is actually for the synchronous pull with lease management doc section. There isn't a sample for real synchronous pull, and there should be.

This task is for updating the GitHub repo for the above. The doc site change is being tracked internally. (134961939)

feywind:
Devsite is also updated, going to go ahead and close this out.

PR #856, opened 2020-01-17
docs(samples): split synchronous pull sample into one with and one without lease management

feywind:
Attached to this issue: https://github.com/googleapis/nodejs-pubsub/issues/855


samples/synchronousPullWithLeaseManagement.js
```diff
@@ -0,0 +1,126 @@
...
+ *
+ * For more information, see the README.md under /pubsub and the documentation
+ * at https://cloud.google.com/pubsub/docs.
+ */
+
+'use strict';
+
+// sample-metadata:
+//   title: Synchronous Pull With Lease Management
+//   description: Receive messages synchronously,
```
JustinBeckwith: The newlines here are interesting.  I wonder if that will have an effect on the generated README?

samples/synchronousPullWithLeaseManagement.js
```diff
@@ -0,0 +1,126 @@
...
+        );
+      }
+    }
+  }
+
+  synchronousPullWithLeaseManagement().catch(console.error);
+  // [END pubsub_subscriber_sync_pull_with_lease_management]
+}
+
+main(...process.argv.slice(2));
```
JustinBeckwith: @bcoe @fhinkel have we come to a conclusion on how we want to handle these yet?

samples/system-test/subscriptions.test.js
```diff
@@ -146,6 +146,17 @@ describe('subscriptions', () => {
     const output = await execPromise(
       `${commandFor('synchronousPull')} ${projectId} ${subscriptionNameOne}`
     );
+    assert.match(output, /Hello/);
+    assert.match(output, /Done./);
+  });
+
+  it('should listen for messages synchronously with lease management', async () => {
+    pubsub.topic(topicNameOne).publish(Buffer.from(`Hello, world!`));
+    const output = await execPromise(
```
JustinBeckwith: So I know it's not part of this PR, but the `execPromise` thing happening here is silly (I think).  Is there a reason for this to be async?  The tests all run serially in their own container.

samples/synchronousPullWithLeaseManagement.js
feywind: I did some of this in the previous PR actually, and it looked like it was okay. synthtool is pulling these lines out as yaml, and that seems to be pretty standard for multi-line values in yaml. I'll double check it.

samples/system-test/subscriptions.test.js
feywind: Hmm. Yeah, it seems like execSync would be okay, but I'm going to guess that this is letting other test running code capture and process outputs while it's running. I'm not sure if that matters. I can try switching it to execSync and see if it causes any issues.

bcoe: `spawnSync` I believe gives more information about the process that ran, such as exit code; might be a good option.

feywind: My hunch above seems to have been correct, that it was execPromise() so that other events could fire, but I also ended up thinking that that was the wrong way to handle the situation. I added an await to the stuff before it and switched to execSync(), and it works fine.

samples/synchronousPullWithLeaseManagement.js
feywind: I re-ran synthtool (I forgot to do that anyway) and it looks good.

samples/synchronousPullWithLeaseManagement.js
feywind: I think this needs to be figured out, but I kinda also think it's a separate issue from this PR (it's just forking an existing sample), so I'm going to leave it as is for now.

Could maybe be handled as part of this issue (or a related one):
https://github.com/googleapis/nodejs-asset/issues/242

feywind:
*blink blink*
```
'node' is not recognized as an internal or external command,
operable program or batch file.
```
Trying again...

Issue #858, opened 2020-01-21
Update client library defaults to match other language libraries for Pub/Sub


feywind:
This is in regards to cps-libs-defaults. We want to update the client libraries to have matching defaults.

feywind:
Turns out there were several more tabs of stuff to update, so I'm re-opening this.

PR #859, opened 2020-01-22
feat(defaults): update defaults for the node client library to match other pub/sub libraries

feywind:
Fixes: https://github.com/googleapis/nodejs-pubsub/issues/858


feywind:
It's just been linter (et al) fixes since Ben's comment, so I'm going to go ahead and merge.

PR #865, opened 2020-01-26
feat(subscription): support oidcToken

stayradiated:
### Issue

It is currently not possible to create a new Subscription with the **Service Account Email** or **Audience** properties configured. These properties are part of the `PushConfig.oidcToken` field which is not being formatted correctly.

### Changes

Similar to the `pushEndpoint` option, when formatting the metadata of a Subscription, the property needs to be moved into the `pushConfig` object.

```typescript
    if (metadata.pushEndpoint) {
      formatted.pushConfig = {
        pushEndpoint: metadata.pushEndpoint,
      };
      delete formatted.pushEndpoint;
    }

    // fix
    if (metadata.oidcToken) {
      formatted.pushConfig = {
        ...formatted.pushConfig,
        oidcToken: metadata.oidcToken,
      };
      delete formatted.oidcToken;
    }
```

I have added tests and updated documentation for `PubSub.createSubscription` and `Subscription.modifyPushConfig`.


feywind:
@bcoe Code review wise this looks fine to me, but I don't know all the inner workings enough to give a yay/nay on the functionality. Do you mind taking a quick look tomorrow? (Or tag someone who would know...)

bcoe:
@feywind this looks reasonable to me too, but I think perhaps we should loop in @bshaffer for a review who's been doing work around ID Tokens.

bcoe:
@stayradiated have you tested end to end with your branch against the actual service?

src/subscription.ts
```diff
@@ -1039,6 +1050,14 @@ export class Subscription extends EventEmitter {
       delete formatted.pushEndpoint;
     }
 
+    if (metadata.oidcToken) {
+      formatted.pushConfig = {
+        ...formatted.pushConfig,
+        oidcToken: metadata.oidcToken,
+      };
+      delete formatted.oidcToken;
```
bcoe: I'm relatively green in this codebase, but mind clarifying why we need to delete the key off the `formatted` object?

bshaffer: It does seem to already be happening above it for `formatted.pushEndpoint` and as well...

feywind: Yeah, that was my take as well, oidcToken is just being treated like what's above. My main question is on the sort of systemic goodness of the patch - I don't understand the gestalt of the library enough to say yay or nay yet.

bshaffer: I believe the changes here just expose access to the underlying API call, and is already exposed in other client libraries. With the exception of deleting the key (is this for memory management?) this is LGTM.

PR #875, opened 2020-02-01
test: modernize mocha config

JustinBeckwith:


feywind:
I think there might actually be an issue here (or uncovered issues from the upgrade). I'll look at this a bit later.

feywind:
Looks like the same failure here as https://github.com/googleapis/nodejs-pubsub/pull/875.

PR #877, opened 2020-02-03
feat: update defaults for batch settings also, and update which result codes will cause a retry

feywind:
Implements the missed parts of https://github.com/googleapis/nodejs-pubsub/issues/858 (batch settings and retry code changes).

### DIFF
--- a/src/publisher/index.ts
+++ b/src/publisher/index.ts
@@ -169,8 +169,8 @@ export class Publisher {
   setOptions(options = {} as PublishOptions): void {
     const defaults = {
       batching: {
-        maxBytes: defaultOptions.publish.maxSize,
-        maxMessages: defaultOptions.publish.maxCount,
+        maxBytes: defaultOptions.publish.maxOutstandingBytes,
+        maxMessages: defaultOptions.publish.maxOutstandingMessages,
         maxMilliseconds: defaultOptions.publish.maxDelayMillis,
       },
       messageOrdering: false,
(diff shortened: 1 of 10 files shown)


src/publisher/index.ts
```diff
@@ -168,9 +169,9 @@ export class Publisher {
   setOptions(options = {} as PublishOptions): void {
     const defaults = {
       batching: {
-        maxBytes: Math.pow(1024, 2) * 5,
-        maxMessages: 1000,
-        maxMilliseconds: 100,
+        maxBytes: defaultOptions.publish.maxOutstandingBytes,
```
JustinBeckwith: If I'm reading this right - the idea was to drop the `* 5` here, right?

src/pull-retry.ts
```diff
@@ -19,15 +19,11 @@ import {StatusObject, status} from '@grpc/grpc-js';
  * retryable status codes
  */
 export const RETRY_CODES: status[] = [
-  status.OK,
```
JustinBeckwith: For TypeScript users, is this technically a breaking change?  Not sure how/if this is exported from `index`.

src/publisher/index.ts
feywind: That's the ultimate effect, yeah. There is a spreadsheet of the various default values in the libraries for different languages, and we're trying to normalize them. The original 5MB value was out of spec for that. I'll paste the link on chat in case that's interesting.

src/pull-retry.ts
feywind: It's a list of what statues might trigger a retry, which got pared down. It's honestly a bit odd that it's even exported. (I removed the export and reran the tests, and they're happy still... but I don't want to do that for fear of it being a real breaking change in case someone's checking it in their own client code.)

PR #878, opened 2020-02-04
chore: first pass at github actions for some of the repo's precommit checks

feywind:
If successful, this will try to shift some of the burden of precommit checks away from Kokoro, to help speed up the process and save us usage quota.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,4 +1,4 @@
-on: [push]
+on: [push, pull_request]
 name: ci
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,48 @@
+on: [push]
```
bcoe: I ended up doing `push` and `pull_request` on https://github.com/googleapis/repo-automation-bots/pull/261/files, do you think we could get away with just `push`?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,48 @@
...
+  license_check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+      - run: npm ci
+      - run: npm run license-check
+  coverage:
```
bcoe: we should add the `test-docs` job too, which tests the links in our reference docs.

.github/workflows/ci.yaml
feywind: `pull_request` might imply `push`, but I'm not sure it does. It probably doesn't hurt to include it.

PR #881, opened 2020-02-07
build: only run push when landing to main branch

bcoe:


feywind:
I did some more thorough changes on another branch. Closing this one so we can do that branch instead.

PR #882, opened 2020-02-07
build: fix ci workflows

feywind:
Fixing a few problems found in the earlier PR:

https://github.com/googleapis/nodejs-pubsub/pull/878

### DIFF
--- a/.kokoro/test.bat
+++ b/.kokoro/test.bat
@@ -21,7 +21,7 @@ cd ..
 @rem we upgrade Node.js in the image:
 SET PATH=%PATH%;/cygdrive/c/Program Files/nodejs/npm
 
-call nvm use v12.14.1
+call nvm use v10
 call which node
 
 call npm install || goto :error
--- a/.kokoro/test.bat
+++ b/.kokoro/test.bat
@@ -23,7 +23,6 @@ SET PATH=%PATH%;/cygdrive/c/Program Files/nodejs/npm
 
 call nvm use v12.14.1
 call which node
-call node --version
 
 call npm install || goto :error
 call npm run test || goto :error
--- a/.kokoro/test.bat
+++ b/.kokoro/test.bat
@@ -21,8 +21,7 @@ cd ..
 @rem we upgrade Node.js in the image:
 SET PATH=%PATH%;/cygdrive/c/Program Files/nodejs/npm
 
-call nvm install lts/dubnium
-call nvm use lts/dubnium
+call nvm use v12.14.1
 call which node
 call node --version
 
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -60,6 +61,7 @@
     "google-auth-library": "^5.5.0",
     "google-gax": "^1.7.5",
     "is-stream-ended": "^0.1.4",
+    "js-green-licenses": "^1.1.0",
```
stephenplusplus: Will this still work as a devDependency?

package.json
```diff
@@ -45,7 +45,8 @@
     "proto:iam": "pbjs -t static-module -w commonjs -p protos google/iam/v1/iam_policy.proto | pbts -o proto/iam.d.ts -",
     "docs-test": "linkinator docs",
     "predocs-test": "npm run docs",
-    "benchwrapper": "node bin/benchwrapper.js"
+    "benchwrapper": "node bin/benchwrapper.js",
+    "license-check": "jsgl --local ."
```
JustinBeckwith: I'd rather not add this or the `js-green-licenses` dependency below, unless we have a plan to do this consistently across the board.  If we *do* want to do this, I'd advocate for a bot over a GitHub Action in this case.

feywind: Ah okay. I can remove it again. I was mostly trying to take the union of all of the GitHub actions I found elsewhere :)

JustinBeckwith: At some point we'll converge on a single config 😆

package.json
feywind: Err... I don't know why that even ended up as a regular dependency. I must've flubbed the command line parameter. Good catch though.

Per Justin's comments I'm probably just removing that anyhow.

package.json
bcoe: I think I'm with @JustinBeckwith on this one, or at least doing the work in a separate PR than this one.

We have our GitHub repos templatized, so once we have a `.github/workflows/ci.yml` that we feel is sufficient, I would actually like to pull it out into our [template directory here](https://github.com/googleapis/synthtool/tree/master/synthtool/gcp/templates/node_library).

.kokoro/test.bat
```diff
@@ -21,8 +21,10 @@ cd ..
 @rem we upgrade Node.js in the image:
 SET PATH=%PATH%;/cygdrive/c/Program Files/nodejs/npm
 
-call nvm use v12.14.1
+call nvm install lts/dubnium
```
JustinBeckwith: what's up with that?

feywind: This is a conversation I had with @bcoe about potentially moving down to a Node 10 version for Windows test, since it seems like the 12 version is failing often. "lts/dubnium" is the most reliable way I found to pull in the latest 10.x. But I am not particularly married to this, I can roll it back.

Issue #905, opened 2020-02-28
Pull subscriber receives messages even after purging!


saharshjain:
#### Environment details
  App engine template 

> service: "my server name"
> runtime: nodejs
> env: flex
> resources:
>   cpu: 2
>   memory_gb: 4
>   disk_size_gb: 10
> network:
>   name: default
> manual_scaling:
>   instances: 1

Package.json

> {
>     "dependencies":{
>         "@google-cloud/pubsub":"1.5.0",
>         "axios":"^0.19.2",
>         "csv-parse":"^4.8.5",
>         "dotenv":"^8.2.0",
>         "google-gax":"1.14.1",
>         "googleapis":"47.0.0",
>         "html-to-text":"^5.1.1",
>         "moment":"^2.24.0",
>         "nylas":"^4.9.0",
>         "path":"^0.12.7",
>         "pg":"^7.18.1",
>         "winston":"^3.0.0"
>     },
>     "devDependencies":{
>         "eslint":"^6.8.0",
>         "eslint-config-airbnb-base":"^14.0.0",
>         "eslint-plugin-import":"^2.20.1"
>     }
> }

#### Steps to reproduce
 I am using GCP App engine to deploy a server 
Operation of server can be understood in the [question](https://stackoverflow.com/questions/60297471/publishing-more-than-actual-messages-in-google-pub-sub-using-node-js-and-csv-par/60306869#60306869)
STR
  1.  Due to some error(fixed by the answer in question) my publisher published 150K message instead of actual ~10K
  2. Now being in production environment I purged them all immediately otherwise users might have got affected. 
3. Even after successful purging my subscriber was able to pull messages more than 1000 messages and process them. 
4. Eventually after an hour of trying I had to delete the subscription and recreate it to stop messages from getting pulled.

Subscription specs

> expiry - never expire
> ack timeout  - 120 seconds
> retain ack messages - yes

Sorry I could not format specs 
![Screenshot 2020-02-28 at 1 25 30 PM](https://user-images.githubusercontent.com/10156548/75521591-16754100-5a2e-11ea-947b-24eb23ed1651.png)
in code style.

feywind:
@saharshjain Thanks for the report! Two questions for you:

1) When you say you purged the old messages, do you mean you clicked the purge button in the GCP control panel page (or something similar)?

[...]

PR #908, opened 2020-03-04
fix: send the ITimestamp protobuf to Pub/Sub for seeking, not JavaScript Date()

feywind:
This fixes a bug in the seek() method that prevented users from passing a date correctly. This is especially relevant for purging (seek to the future).

May relate: https://github.com/googleapis/nodejs-pubsub/issues/905
Related internally: 150712874

### DIFF
--- a/src/subscription.ts
+++ b/src/subscription.ts
@@ -895,7 +895,7 @@ export class Subscription extends EventEmitter {
       const dateMillis = (snapshot as Date).getTime();
       reqOpts.time = {
         seconds: Math.floor(dateMillis / 1000),
-        nanos: Math.floor(dateMillis % 1000) * 1000
+        nanos: Math.floor(dateMillis % 1000) * 1000,
       };
     } else {
       throw new Error('Either a snapshot name or Date is needed to seek to.');
--- a/src/subscription.ts
+++ b/src/subscription.ts
@@ -892,7 +892,11 @@ export class Subscription extends EventEmitter {
     if (typeof snapshot === 'string') {
       reqOpts.snapshot = Snapshot.formatName_(this.pubsub.projectId, snapshot);
     } else if (Object.prototype.toString.call(snapshot) === '[object Date]') {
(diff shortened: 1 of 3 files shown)


src/subscription.ts
```diff
@@ -892,7 +892,11 @@ export class Subscription extends EventEmitter {
     if (typeof snapshot === 'string') {
       reqOpts.snapshot = Snapshot.formatName_(this.pubsub.projectId, snapshot);
     } else if (Object.prototype.toString.call(snapshot) === '[object Date]') {
-      reqOpts.time = snapshot as google.protobuf.ITimestamp;
+      const dateMillis = (snapshot as Date).getTime();
+      reqOpts.time = {
+        seconds: Math.floor(dateMillis / 1000),
```
bcoe: Part of me thinks [`process.hrtime()`](https://nodejs.org/api/process.html#process_process_hrtime_time) would read better, at the same time it makes things harder to webpack ... so I lean towards your approach.

system-test/pubsub.ts
```diff
@@ -862,6 +862,30 @@ describe('pubsub', () => {
...
         });
       });
+
+      it('should seek to a future date (purge)', done => {
+        const testText = 'Oh no!';
+
+        // Forward-seek to remove any messages from the queue (those were
+        // placed there in before()).
+        //
+        // We... probably won't be using this in 3000?
```
bcoe: I look forward to someone at a hover desk addressing this bug.

src/subscription.ts
feywind: Hmm, I didn't know about hrtime(). That's pretty nice. I think I'll go ahead and keep it as it is though, for the reason you gave. I don't think we support using the libraries from a browser anyway, right? But it doesn't seem good to introduce future problems just in case.

system-test/pubsub.ts
feywind: <Marty McFly meme>

PR #912, opened 2020-03-06
build(tests): fix coveralls and enable build cop

bcoe:


feywind:
Something weird is happening here, and these tests seem to fail locally too, so I'm just going to come back to this.

```
  1) topics
       should create a topic:
     Uncaught DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
...
  2) topics
       should create a topic:
     Error: done() called multiple times
      at process._tickCallback (internal/process/next_tick.js:68:7)
```

There's only one `should create a topic`.

Issue #817, opened 2019-11-12
Publisher Client `close()` method


colmsnowplow:
**Is your feature request related to a problem? Please describe.**

There's currently no method to close a client. I'm using the client in a persistent environment, which isn't guaranteed to be actively processing data at any given time. 

If there's a delay between messages, there's a chance that the network connection fails to resolve, eg. if the connection is closed on the server side but not client side. 

In order to cover this, I need to open a new connection and I have no way to close the stale one, so basically this eventually leads to a proliferation of open but unused sockets.

 **Describe the solution you'd like**

A `client.close()` method which allows me to terminate a client connection client-side.

 **Describe alternatives you've considered**

Manually closing sockets or terminating the entire environment every time we hit a stale connection are the only real options I can see - neither are desirable for obvious reasons.

merlinnot:
This is a great feature request, I'd benefit from it too.

There's a similar FR against Firestore: https://github.com/googleapis/nodejs-firestore/issues/769
And a preliminary PR against API Client Generator for TypeScript: https://github.com/googleapis/gapic-generator-typescript/pull/126

merlinnot:
@bcoe @callmehiphop It was now implemented in the Firestore client and I got impressive gains thanks to it, see https://github.com/googleapis/nodejs-firestore/issues/769#issuecomment-568165126. Is there a chance to prioritize this feature somehow? I have the exact same issue with PubSub, as I had with Firestore.

bcoe:
@merlinnot just added to the schedule for our next library group meeting; @feywind probably worth having this on your radar 👍

colmsnowplow:
Good to hear this will be discussed! 

For the record/in case it's pertinent I read over the issue @merlinnot linked to for firestore, and we face exactly the same problem (although he described it better than I!)

merlinnot:
@bcoe Did the meeting happen already?

feywind:
@merlinnot @colmsnowplow It's been discussed a couple of times. I haven't heard a super definite "yay do it", but I think the consensus is that we're planning to do it. Basically that it probably shouldn't be necessary to use a close() method, but that it will still probably help people for now. It's behind another fire or two though :)

merlinnot:
[...]

PR #916, opened 2020-03-18
feat: add a close() method to PubSub, and a flush() method to Topic/Publisher

feywind:
Fixes: https://github.com/googleapis/nodejs-pubsub/issues/817

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,7 +11,7 @@ jobs:
       matrix:
         node: [8, 10, 12, 13]
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
         with:
           node-version: ${{ matrix.node }}
@@ -30,7 +30,7 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
         with:
           node-version: 12
@@ -39,7 +39,7 @@ jobs:
   docs:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
         with:
           node-version: 12
@@ -48,7 +48,7 @@ jobs:
   coverage:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
         with:
           node-version: 12
--- a/src/publisher/index.ts
+++ b/src/publisher/index.ts
@@ -14,14 +14,14 @@
  * limitations under the License.
  */
 
-import {promisifyAll} from '@google-cloud/promisify';
+import {promisify, promisifyAll} from '@google-cloud/promisify';
 import * as extend from 'extend';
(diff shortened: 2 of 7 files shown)


feywind:
Looks like there's an issue in checkout@v1, and we hadn't updated all of the CI actions to v2.
https://github.com/actions/checkout/issues/23

src/pubsub.ts
```diff
@@ -281,6 +282,33 @@ export class PubSub {
...
+   */
+  close(callback?: EmptyCallback): Promise<void> | void {
+    const definedCallback = callback || (() => {});
+    if (this.isOpen) {
+      this.closeAllClients_()
+        .then(() => {
+          definedCallback(null);
+        })
+        .catch(definedCallback);
+      this.isOpen = false;
```
bcoe: I might move this to above `this.closeAllClients_`, just so it's clear that `this.isOpen` is already `false`, before the promise chain finishes executing.

test/pubsub.ts
```diff
@@ -243,6 +251,15 @@ describe('PubSub', () => {
       assert.deepStrictEqual(pubsub.api, {});
     });
 
+    it('should default to the opened state', () => {
+      assert.strictEqual(pubsub.isOpen, true);
+    });
+
+    it('should not be in the opened state after close()', async () => {
```
bcoe: might be tempted to flesh this test out a bit more, checking that the internal state reflects that things were closed, as an example has the gRPC stream actually had close called on it? has `closeAllClients_` resovled.

.github/workflows/ci.yaml
```diff
@@ -11,7 +11,7 @@ jobs:
       matrix:
         node: [8, 10, 12, 13]
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
```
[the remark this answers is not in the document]
feywind: I do agree actually, but the CI tests weren't passing :) so I just went ahead and threw this in.

src/publisher/index.ts
```diff
@@ -72,6 +72,34 @@ export class Publisher {
...
+  flush(callback?: EmptyCallback): Promise<void> | void {
+    const definedCallback = callback ? callback : () => {};
+
+    const publishes = [promisify(this.queue.publish)()];
+    Array.from(this.orderedQueues.values()).forEach(q =>
+      publishes.push(promisify(q.publish)())
+    );
+    const allPublishes = Promise.all(publishes);
+
+    allPublishes
```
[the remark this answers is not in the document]
feywind: Yeah, I about 350% agree... unfortunately the current code is mostly callback-based, and a few event-based bits, so pulling the async/await far up the stack was looking difficult. I'd really like to go back and do a pass of just converting everything to straightforward async/await/Promises, but I don't want to do that as part of this.

src/pubsub.ts
feywind: Good call, the whole thing should stop working if someone calls close() at all.

test/pubsub.ts
feywind: This test does look really minimal, but I think those points are covered elsewhere, actually. e.g. this shouldn't resolve until `closeAllClients_` resolves:
```
      await pubsub.close();
```

There's a separate test that stubs out `close` for the gax clients in `closeAllClients_`, so that should also be covered.

feywind:
I added a note about the potentially flaky test, re-running Kokoro for now.

PR #917, opened 2020-03-19
chore: remove snippet leading whitespace

chingor13:


feywind:
The system test failure looks like it might just be a flaky test that should be linked to that tracking issue (will do in a sec). The docs failure I'm getting in my unrelated PR also. It seems like we might've changed something lately that broke this, but I don't know why it wouldn't have shown up in the PR for that change.

Issue #937, opened 2020-03-31
PubSub.close throws error


mad-it:
This bug report is related to https://github.com/googleapis/nodejs-pubsub/pull/916.

#### Environment details

  - OS: Linux
  - Node.js version: v10.16.3
  - npm version:  6.9.0
  - `@google-cloud/pubsub` version: 1.7.0

#### Steps to reproduce

Here is a test written to test this newly introduced `close` method.

```
import { PubSub } from '@google-cloud/pubsub';

it('test pubsub close', async (): Promise<void> => {
  expect.assertions(1);
  const topicName = 'test-topic';

  const pubSub = new PubSub();

  const [topic] = await pubSub.topic(topicName).get({ autoCreate: true });

  await topic.publish(Buffer.from('test-message'));

  expect(await pubSub.topic(topicName).exists()).toStrictEqual([true]);

  await pubSub.close();
});

```

When running this simple test in Jest, I get the following error.

```
 FAIL  src/test.spec.ts (7.801s)
  ✕ test pubsub close (106ms)

  ● test pubsub close

    TypeError: gaxClient.close is not a function

      13 |   expect(await pubSub.topic(topicName).exists()).toStrictEqual([true]);
      14 | 
    > 15 |   await pubSub.close();
         |                ^
      16 | });
      17 | 

      at PubSub.closeAllClients_ (node_modules/@google-cloud/pubsub/src/pubsub.ts:983:31)
      at PubSub.close (node_modules/@google-cloud/pubsub/src/pubsub.ts:302:12)
      at PromiseCtor (node_modules/@google-cloud/promisify/build/src/index.js:69:28)
      at PubSub.wrapper (node_modules/@google-cloud/promisify/build/src/index.js:54:16)
      at Object.it (src/test.spec.ts:15:16)
```

stephenplusplus:
// @feywind

feywind:
@mad-it @stephenplusplus Thanks! I will take a look in a moment.

feywind:
Well, this will teach me to not write a system test for a feature. :S The Pub/Sub client API objects are hand-written and not using a standard generated gax client, and this gets cast to a standard gax interface, which is really confusing. I'll work on getting this piped through tomorrow and add a system test based on the case above.

Thanks again for the report!

feywind:
I'm not sure I like the issue auto-close. Anyway, this PR should fix the problem, can you try again?
https://github.com/googleapis/nodejs-pubsub/pull/941

[...]

PR #923, opened 2020-03-23
build!: convert to typescript

summer-ji-eng:
BREAKING CHANGE：` topicPath` has been changed to `projectTopicPath`

Some synthtool hack will be removed once pubsub upgrade to `gts:2.0.0`

PR #948, opened 2020-04-08
fix: add a close() method to the generated SubscriberClient for 1.x

feywind:
Apparently the gapic generated SubscriberClient also doesn't have a `close()` method. This adds another synth.py hack and corresponding change in the .js file. Also a new system-test to catch subscribers specifically.

Follow-on fix for https://github.com/googleapis/nodejs-pubsub/issues/937

A proper fix is happening at https://github.com/googleapis/nodejs-pubsub/pull/923 but it will require bumping major versions, so I'm basically backporting this for 1.7.x.

### DIFF
--- a/synth.py
+++ b/synth.py
@@ -62,6 +62,26 @@ s.replace("src/v1/publisher_client.js",
           "    // note: editing generated code\n"
           "    this.publisherStub = publisherStub;\n"
           "    \g<0>")
+s.replace("src/v1/subscriber_client.js",
+          "   \* Parse the projectName from a project resource.\n",
+          "   * Terminate the GRPC channel and close the client.\n"
+          "   * note: editing generated code\n"
+          "   *\n"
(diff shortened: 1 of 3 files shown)


synth.py
```diff
@@ -62,6 +62,26 @@
...
           "    this.publisherStub = publisherStub;\n"
           "    \g<0>")
+s.replace("src/v1/subscriber_client.js",
+          "   \* Parse the projectName from a project resource.\n",
+          "   * Terminate the GRPC channel and close the client.\n"
+          "   * note: editing generated code\n"
+          "   *\n"
+          "   * The client will no longer be usable and all future behavior is undefined.\n"
+          "   */\n"
+          "  close() {\n"
```
bcoe, line 72: gross, but if it does the trick 😄

feywind: I know, right? =_=

PR #952, opened 2020-04-09
fix(close): ensure in-flight messages are drained

### LINKED ISSUE #951: Race condition when flushing message queues while a batch is in-flight
There seems to be a race condition in `subscriber.close()` (`_waitFlush`) when the check for pending requests is made while a batch is currently in-flight.

I think we should track the in-flight requests as well, and resolve the `_waitFlush` when in-flight requests have drained.

#### Environment details

  - OS: macOS Catalina
  - Node.js version: 13.8.0
  - npm version: 6.14.3
  - `@google-cloud/pubsub` version: 1.7.1

#### Steps to reproduce

  1. Add a topic and subscription.
  2. Subscribe and `ack` a message
  3. Immediately after `ack`ing (`message.ack(): void`), `await subscription.close()`
  4. Delete the subscription.
  5. Depending on timing, you will see something like `Cannot ack messages: Subscription does not exist`

I used this test to reproduce the issue, and also tested a potential fix (PR incoming):

```ts
import { getEnv } from '../../../env/env'
import * as pubsub from '@google-cloud/pubsub'
import { v4 } from 'uuid'
import { defer } from 'bluebird'

jest.setTimeout(1000000)

test('errors if messages are inflight while closing, then deleting subscription causing it to fail', async () => {
  const client = createPubSubSdk(getEnv())
  // with enough iterations we should be able to trigger it.
  // NOTE: when the fix is applied, this succeeds.
  for (let i = 0; i < 100; i++) {
    const topicName = `bug-topic-${v4()}`
    const subName = `bug-sub-${v4()}`
    const [topic] = await client.createTopic(topicName)
    const [sub] = await topic.createSubscription(subName, {
      expirationPolicy: { ttl: { seconds: 60 } },
    })

    await topic.publish(Buffer.from(JSON.stringify({ hello: 'world' })))

    const onHandlerEnter = defer()
    const onError = defer()
    sub.setOptions({ batching: { maxMilliseconds: 100 } })
    sub.addListener('message', (message: pubsub.Message) => {
      message.ack()
      onHandlerEnter.resolve()
    })
    sub.addListener('error', (err) => {
      console.log(err)
      onError.reject(err)
    })
    sub.open()

    await onHandlerEnter.promise

    // Kicks off the flush
    await new Promise((resolve) => setTimeout(resolve, 100))

    // Close while requests are pending. Wait for it.
    await sub.close()
    // Delete the subscription. This will trigger an error in ack.
    await sub.delete()
    await topic.delete()
    await Promise.race([
      new Promise((resolve) => setTimeout(resolve, 1000)),
      onError.promise,
    ])
  }
})

function createPubSubSdk(env: any) {
  return new pubsub.PubSub({
    projectId: env.GCLOUD_PROJECT,
    credentials: JSON.parse(env.GCLOUD_CREDENTIALS),
  })
}
```

jeffijoe:
Track in-flight requests and add an `onDrain` deferred that resolves when all in-flight `ack`/`modAcks` are done.

Fixes #951


jeffijoe:
@googlebot I signed it!

feywind:
Thanks for the contribution, @jeffijoe ! I'll review in just a sec.

test/message-queues.ts
```diff
@@ -209,6 +210,40 @@ describe('MessageQueues', () => {
         setImmediate(() => messageQueue.flush());
         return promise;
       });
+
+      it('should resolve onDrain only after all in-flight messages have been flushed', async () => {
+        const patchedQueue = new MessageQueue(subscriber);
+        const origSendBatch = patchedQueue._sendBatch;
+        const log: string[] = [];
+        const sendDone = defer();
+        patchedQueue._sendBatch = async function _sendBatch(
```
feywind, line 219: I wonder if this might be done using `sandbox.stub` just to be consistent with the rest of the tests. I'm going to ask someone who knows a bit more about Sinon than me.

jeffijoe: Do you mean just the patching of the method?

feywind: Yep! Sorry for not being clear there. On second thought, I think it's not a big deal because you're making your own `MessageQueue` anyway.

jeffijoe: I wanted to stay consistent with the rest of the source but I was not sure how else I would pull this off. Do you have any suggestions?

jeffijoe:
If you are wondering about the force-pushes, I am just rebasing to latest `master`.

test/message-queues.ts
feywind: I think you could probably use the sandbox on the existing `messageQueue._sendBatch`, if you wanted to be a bit more consistent with the other tests. That might be nice if we change something about how the "global" `MessageQueue` is wired up under `describe` later, but I don't think it's worth spending much time on.

feywind:
Yep, no worries on the rebase!

test/message-queues.ts
jeffijoe: I changed it and rebased to tip of master but now other unrelated tests are failing. Is master healthy right now?

feywind: It _should_ be, but we have been tracking down some flaky tests for a bit, too. I'll look.

feywind:
I think the rebase is actually confusing our CI bots, so you might want to just use merge to get the tests running automatically.

jeffijoe:
Kokoro seemed to have issues even with my first non-rebased commit though? I can push a dummy commit if you want and see if it resolves anything?

feywind:
Sure, let's try it for science :)

It looks like the checks are passing at this point. Hopefully we'll have our automated sample failure bot going on this repo soon.

src/message-queues.ts
```diff
@@ -142,9 +146,15 @@ export abstract class MessageQueue {
       this._subscriber.emit('error', e);
     }
 
+    this.numInFlightRequests -= batchSize;
     if (deferred) {
       deferred.resolve();
     }
+
+    if (this.numInFlightRequests === 0 && this._onDrain) {
```
feywind, line 154: Oh, super minor thing here that I thought @bcoe was going to comment on - to be overly defensive, it might be nice to make this `<= 0`.

feywind:
Apparently the actual issue with the CI here is that Kokoro is having issues. I'll try to get this merged tomorrow.

src/message-queues.ts
jeffijoe: Done! :)

PR #964, opened 2020-04-15
chore: clean up synthtool

summer-ji-eng:


synth.py
```diff
@@ -20,7 +20,8 @@
         'package-name': f'@google-cloud/pubsub',
         'main-service': f'pubsub',
         'bundle-config': f'google/pubsub/{version}/pubsub_gapic.yaml',
-        'template': f'typescript_gapic'
+        'template': f'typescript_gapic',
```
alexander-fenster, line 23: nit: here and above, `f'...'` means "formatted string" - the string where you can use `{placeholders}` to insert a variable. It's actually only used in `bundle-config`, but in `package-name`, `main-service`, and `template` there should be no `f'...'`, just regular strings.

synth.py
```diff
@@ -20,7 +20,8 @@
...
+        "iam-service": "true"
```
alexander-fenster, line 24: super nit: the script above uses single quotes, not double quotes :)

synth.py
summer-ji-eng: gotcha. thanks

summer-ji-eng:
> LGTM only with one question, the `// eslint-disable-next-line @typescript-eslint/no-unused-vars` in test files are auto-generated or manually added? If it's manual work we may want to do it through synth.py to avoid lint error. Thanks!

I guess the test you mention here is manual test not the test generated from micro-generator.

PR #994, opened 2020-05-07
refactor: use grpc from google-gax

alexander-fenster:
Getting rid of any direct dependency on gRPC.

~~Also, `grpc` dev dependency looks completely unused, so removing it :)~~
Upd. It's used by `bin/benchwrapper.js`. Changed that to use gRPC exported from `google-gax` as well.

### DIFF
--- a/bin/benchwrapper.js
+++ b/bin/benchwrapper.js
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-const grpc = require('grpc');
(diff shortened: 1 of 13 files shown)


alexander-fenster:
@JustinBeckwith Yeah. We could reexport `status` and others, or we could do `const status = grpc.status`, but I'm not a fan of all this :)

JustinBeckwith:
Heh, nah it's all good - this is more explicit

Issue #825, opened 2019-11-20
Error [ERR_STREAM_WRITE_AFTER_END]: write after end


glesage:
#### Environment details

  - OS: Google Cloud Container-Optimized OS
  - Node.js version: 10.17.0
  - npm version: 6.12.0
  - `@google-cloud/pubsub` version: 1.x (1.1.5, tried reverting down to 1.0.0, same error, currently using 0.21.1)

#### Steps to reproduce

Unclear so far but have been seeing it after publishing 10k+ messages to various topics. Will get back with more details when we can but if anyone else is seeing this to confirm this isn't a config issue let us know.

We are using our own wrapper around pubsub which you can find here: https://github.com/starchup/google-pubsub-wrapper, extremely bare bones

#### Error stacktrace

```

E 2019-11-19T16:49:45.701927239Z events.js:174
E 2019-11-19T16:49:45.702039214Z       throw er; // Unhandled 'error' event
E 2019-11-19T16:49:45.702047675Z       ^
E 2019-11-19T16:49:45.702054717Z 
E 2019-11-19T16:49:45.702060707Z Error [ERR_STREAM_WRITE_AFTER_END]: write after end
E 2019-11-19T16:49:45.702068184Z     at writeAfterEnd (_stream_writable.js:248:12)
E 2019-11-19T16:49:45.702074930Z     at ClientHttp2Stream.Writable.write (_stream_writable.js:296:5)
E 2019-11-19T16:49:45.702081282Z     at filterStack.sendMessage.then.message (/node_modules/@grpc/grpc-js/build/src/call-stream.js:372:34)
E 2019-11-19T16:49:45.702088817Z     at process._tickCallback (internal/process/next_tick.js:68:7)
E 2019-11-19T16:49:45.702094878Z Emitted 'error' event at:
E 2019-11-19T16:49:45.702121372Z     at errorOrDestroy (internal/streams/destroy.js:107:12)
E 2019-11-19T16:49:45.702164237Z     at onwriteError (_stream_writable.js:436:5)
E 2019-11-19T16:49:45.702169699Z     at onwrite (_stream_writable.js:461:5)
E 2019-11-19T16:49:45.702174990Z     at /node_modules/async-listener/glue.js:188:31
E 2019-11-19T16:49:45.702180323Z     at process._tickCallback (internal/process/next_tick.js:63:19)
```
Thanks!

callmehiphop:
[...]

PR #1030, opened 2020-06-12
feat: explicitly pull in grpc-js ~1.0.5

feywind:
Adds a direct dependency to grpc-js so we can pull in versions faster than gax, if needed. This is also correct (imo) because the Pub/Sub library is already directly referencing things in grpc-js.

Also bump us to ~1.0.5 so we get the fix for the stream write after end. Hopefully fixes https://github.com/googleapis/nodejs-pubsub/issues/825.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,8 @@
     "is-stream-ended": "^0.1.4",
     "lodash.snakecase": "^4.1.1",
     "p-defer": "^3.0.0",
-    "protobufjs": "^6.8.1"
+    "protobufjs": "^6.8.1",
+    "@grpc/grpc-js": "~1.0.5"
   },
   "devDependencies": {
     "@grpc/proto-loader": "^0.5.4",


alexander-fenster:
I don't feel too positive or too negative about that - there are pros and cons :)

feywind:
After talking it over internally, I think I'm going to pause this and just put in a gax PR, and repurpose this one for pulling in the new gax.

Sytten:
I hit a bug when I updated to v2: `../../node_modules/@google-cloud/pubsub/build/src/index.d.ts(77,30): error TS2307: Cannot find module '@grpc/grpc-js' or its corresponding type declarations.` Is this related?

feywind:
Gax versions were updated by https://github.com/googleapis/nodejs-pubsub/pull/1085 so I'm going to go ahead and close this.

PR #1032, opened 2020-06-16
feat: support for detaching subscriptions

### LINKED ISSUE #1031: Add support for detaching subscriptions
The server API is not released yet, but we need to add client support for detaching subscriptions, as well as the "detached" flag for subscription metadata.

https://github.com/googleapis/googleapis/blob/e45a2f54dce1ac68497acf305145f9e5bb7e1e48/google/pubsub/v1/pubsub.proto#L130

feywind:
This is not yet available for public service use, but client library support is in progress.

PR is a work in progress.

Fixes: https://github.com/googleapis/nodejs-pubsub/issues/1031


samples/detachSubscription.js
```diff
@@ -0,0 +1,64 @@
...
+      `Subscription ${subscriptionName} detached status: ${metadata.detached}`
+    );
+
+    await sub.detach();
+    console.log(`Subscription ${subscriptionName} detached.`);
+
+    const updatedMetadata = await sub.getMetadata()[1];
+    console.log(
+      `Subscription ${subscriptionName} detached status: ${updatedMetadata.detached}`
+    );
```
anguillanneuf, line 57: How about 1. call `.detach()` 2. if detached, print out successfully detached, else not detached?

samples/detachSubscription.js
```diff
@@ -0,0 +1,64 @@
...
+//   title: Detach Subscription
+//   description: Detaches a subscription from a topic.
+//   usage: node detachSubscription.js <subscription-name>
+
+function main(subscriptionName = 'YOUR_SUBSCRIPTION_NAME') {
+  // [START pubsub_detach_subscription]
+  /**
+   * TODO(developer): Uncomment these variables before running the sample.
+   */
+  // const subscriptionName = 'YOUR_SUBSCRIPTION_NAME';
```
anguillanneuf, line 35: Add a comment - provide an existing subscription.

samples/detachSubscription.js
feywind: Yeah, what I was trying to demonstrate here is the ability to check the "detached" flag from metadata. But it seems like both are useful. And I do want to add a helper that does the metadata check, because it looks gross as it is.

feywind:
Seems like codecov is annoyed because the system-test doesn't cover the new sample. (Which is true, but I was just planning to make a second PR that we can merge when it's ready.)

feywind:
@anguillanneuf @pradn Do y'all mind giving this a look-over?

samples/detachSubscription.js
```diff
@@ -0,0 +1,71 @@
...
+// limitations under the License.
+
+/**
+ * This sample demonstrates how to detach subscriptions with the
+ * Google Cloud Pub/Sub API.
+ *
+ * For more information, see the README.md under /pubsub and the documentation
+ * at https://cloud.google.com/pubsub/docs.
+ *
+ * Note: This sample is NOT currently included in the system-test, because
```
JustinBeckwith, line 22: Should we put a `DO NOT MERGE` on this until that's fixed?

samples/detachSubscription.js
```diff
@@ -0,0 +1,71 @@
...
+    console.log(
+      `Subscription ${subscriptionName} 'after' detached status: ${updatedDetached}`
+    );
+  }
+
+  detachSubscription();
+  // [END pubsub_detach_subscription]
+}
+
+process.on('unhandledRejection', err => {
```
JustinBeckwith, line 67: I believe this is the stuff we're looking for now:
```js

main(...process.argv.slice(2)).catch(err => {
  console.error(err.message);
  process.exitCode = 1;
});
process.on('unhandledRejection', (err) => {
  console.error(err.message);
  process.exitCode = 1;
});
```

src/pubsub.ts
```diff
@@ -541,32 +546,125 @@ export class PubSub {
...
   }
+
+  detachSubscription(
+    name: string,
+    gaxOpts?: CallOptions
+  ): Promise<DetachSubscriptionResponse>;
+  detachSubscription(name: string, callback: DetachSubscriptionCallback): void;
+  detachSubscription(
+    name: string,
+    gaxOpts: CallOptions,
```
JustinBeckwith, line 557: Usually if `gaxOpts` is optional in one invocation model (promises), it can be optional in the callback model as well.  little weird that you MUST specify `gaxOpts` only for callbacks.

Side crazy question - what if we just like didn't support callbacks for new APIs? May be a good way to try this out and measure the response.

src/pubsub.ts
```diff
@@ -541,32 +546,125 @@ export class PubSub {
...
+   * @param {object} [gaxOpts] Request configuration options, outlined
+   *     here: https://googleapis.github.io/gax-nodejs/interfaces/CallOptions.html.
+   * @param {DetachSubscriptionCallback} [callback] Callback function.
+   * @returns {Promise<DetachSubscriptionResponse>}
+   *
+   * @example
+   * const {PubSub} = require('@google-cloud/pubsub');
+   * const pubsub = new PubSub();
+   *
+   * pubsub.detachSubscription('my-sub', function(err, topic, apiResponse) {
```
JustinBeckwith, line 575: Maybe use an arrow function?

src/pubsub.ts
```diff
@@ -541,32 +546,125 @@ export class PubSub {
...
+        this.options.port = undefined;
+      }
+    } else {
+      this.options.port = parseInt(baseUrlParts[1], 10);
+    }
+
+    // If this looks like a GCP URL of some kind, don't go into emulator
+    // mode. Otherwise, supply a fake SSL provider so a real cert isn't
+    // required for running the emulator.
+    const officialUrlMatch = this.options.servicePath!.match(
```
JustinBeckwith, line 660: `this.options.servicePath!.endsWith('.googleapis.com')` is so readable for regex-impaired folk like me

src/pubsub.ts
```diff
@@ -541,32 +546,125 @@ export class PubSub {
...
+      callback!
+    );
+  }
+
   /**
    * Determine the appropriate endpoint to use for API requests, first trying
-   * the local `apiEndpoint` parameter. If the `apiEndpoint` parameter is null
-   * we try Pub/Sub emulator environment variable (PUBSUB_EMULATOR_HOST),
-   * otherwise the default JSON API.
+   * the local `apiEndpoint` parameter. If the `apiEndpoint` parameter is null,
```
JustinBeckwith, line 614: This whole section below feels disconnected from the rest of the content of the PR.  Should these be two separate PRs?

feywind: Fair. I needed this to test the rest of the PR, but this part could be merged first.

src/pubsub.ts
feywind: I was mostly trying to match existing examples in the comments, but I do agree it should be an arrow function.

src/pubsub.ts
feywind: Yeah, I was again just matching existing style. There was some discussion about dropping callbacks earlier, and I think that was unpopular. I'm up for discussing it though.

samples/detachSubscription.js
feywind: So, I'm slightly concerned there is Zalgo in that ordering. (This objection is a bit ironic since I helped author that sample style update. >_>) That can be a separate discussion though.

samples/detachSubscription.js
feywind: The other languages seem to be doing that, so I might shift that way, yeah.

samples/detachSubscription.js
JustinBeckwith: Ugh.  @bcoe @sofisl @fhinkel

samples/detachSubscription.js
```diff
@@ -0,0 +1,71 @@
...
+
+  async function detachSubscription() {
+    // Gets the status of the existing subscription
+    const sub = pubSubClient.subscription(subscriptionName);
+    const [detached] = await sub.detached();
+    console.log(
+      `Subscription ${subscriptionName} 'before' detached status: ${detached}`
+    );
+
+    await pubSubClient.detachSubscription(subscriptionName);
```
anguillanneuf, line 54: Does `detachSubscription` throw explicit errors?  If so, we could list them in the sample.

feywind: Do you mean things like "the subscription was already detached"? I'm actually not sure.

All of those gRPC calls are theoretically capable of throwing a bunch of gRPC errors, and the other samples don't call those out, so I'd want to keep everything as harmonized across samples there as I can.

samples/detachSubscription.js
feywind: Updating this from internal discussions: the feature will be available to allow-listed test projects, so we can go ahead with merging the full set in. I'll re-enable this.

src/pubsub.ts
feywind: Hmm, I see what was going on here:
```
  detachSubscription(
    name: string,
    gaxOpts?: CallOptions
  ): Promise<DetachSubscriptionResponse>;
  detachSubscription(name: string, callback: DetachSubscriptionCallback): void;
  detachSubscription(
    name: string,
    gaxOpts: CallOptions,
    callback: DetachSubscriptionCallback
  ): void;
  detachSubscription(
    name: string,
    optsOrCallback?: CallOptions | DetachSubscriptionCallback,
    callback?: DetachSubscriptionCallback
  ): Promise<DetachSubscriptionResponse> | void {
```

Basically this is just TypeScript resolution shenanigans. If you call with only gaxOpts, you want it to give you a Promise. If you call with gaxOpts and a callback, you get the second variant. If you call with only a callback, it will fall through to the last one, which has optsOrCallback. I don't think it has an effect one way or another on the runtime behavior, but it's sort of gross... but I get why Dave did it that way, too.

My "want a pony" goals for the upcoming quarters include switching everything to use Promise internally, and wrap stuff with callbackify(). There are a few things I need to figure out first though. (Like methods that just pass through callbacks directly.)

samples/detachSubscription.js
anguillanneuf: I was referring to explicit exceptions. Java/Python samples are generally against generic exceptions so I was wondering about Node.

```python
    try:
        publisher_client.detach_subscription(subscription_path)
    except (GoogleAPICallError, RetryError, ValueError, Exception) as err:
        print(err)
```

feywind: Ah, I see. Yeah, the samples are all written in JavaScript, rather than TypeScript, so there's basically no rigor anywhere in them, outside of catching exceptions. I don't think even TypeScript has any real concept of typed exceptions. You just basically have to catch all the exceptions and filter for what you want. We could put a comment on it showing what exceptions might happen, but it would be very hard to figure that out for sure, and I'm afraid of it getting out of sync (e.g. if gRPC changed their exceptions thrown, and the Pub/Sub samples don't get updated).

feywind:
The system test runs locally for me, and the log file for Kokoro seems to be blank. But this reminds me that this PR probably also needs a system test, anyway.

samples/detachSubscription.js
anguillanneuf: Thanks for explaining. I see now.

anguillanneuf (approved):
LGTM samples :)

PR #1060, opened 2020-07-16
fix: properly handle non-emulator alternate endpoints for pub/sub

feywind:
Broken off from https://github.com/googleapis/nodejs-pubsub/pull/1032
Related to: https://github.com/googleapis/nodejs-pubsub/issues/1031

The client library currently does not properly handle alternate apiEndpoint configuration values. They are assumed to be an emulator, but they could e.g. be a staging server or regional endpoint. This corrects that behavior.


feywind:
@JustinBeckwith I broke this out into its own PR, and added the changes you suggested over there.

anguillanneuf:
@feywind Thanks for the fix! Is support for alternate endpoints available for other product emulator (BigTable/Spanner/etc.) other Pub/Sub libraries?

src/pubsub.ts
```diff
@@ -543,30 +543,60 @@ export class PubSub {
...
   /**
    * Determine the appropriate endpoint to use for API requests, first trying
-   * the local `apiEndpoint` parameter. If the `apiEndpoint` parameter is null
-   * we try Pub/Sub emulator environment variable (PUBSUB_EMULATOR_HOST),
-   * otherwise the default JSON API.
+   * the local `apiEndpoint` parameter. If the `apiEndpoint` parameter is null,
+   * we try the standard `gcloud alpha pubsub` environment variable
+   * (CLOUDSDK_API_ENDPOINT_OVERRIDES_PUBSUB), and failing that, we try the
+   * Pub/Sub emulator environment variable (PUBSUB_EMULATOR_HOST).
+   * Otherwise the default production API is used.
```
anguillanneuf, line 550: You probably also want to update https://github.com/googleapis/nodejs-pubsub/blob/a5093c439e5cb13a9d9152d0483373206634ac9b/src/pubsub.ts#L187-L190

JustinBeckwith (approved):

src/pubsub.ts
```diff
@@ -543,30 +543,60 @@ export class PubSub {
...
-    if (!apiEndpoint && !process.env.PUBSUB_EMULATOR_HOST) {
+    // We allow an override from the client object options, or from
+    // one of these variables. The CLOUDSDK variable is provided for
+    // compatibility with the `gcloud alpha` utility.
+    const gcloudVarName = 'CLOUDSDK_API_ENDPOINT_OVERRIDES_PUBSUB';
+    const emulatorVarName = 'PUBSUB_EMULATOR_HOST';
+    const apiEndpoint =
+      this.options.apiEndpoint ||
+      process.env[gcloudVarName] ||
+      process.env[emulatorVarName];
```
stephenplusplus, line 563: This could be backwards, but I would think "PUBSUB_EMULATOR_HOST" would be before "CLOUDSDK_API_ENDPOINT_OVERRIDES_PUBSUB" in precedence, as "CLOUDSDK" sounds like it's set automatically in that environment, compared to "PUBSUB_EMULATOR_HOST" which is deliberately set by the user. I suppose it's a rare scenario where both are defined, but we've always treated "PUBSUB_EMULATOR_HOST" as something a user would define as an alternative to "options.apiEndpoint".

feywind:
@anguillanneuf The last word I heard is that we are working on normalizing these environment variables. So I wouldn't expect a lot of consistency just yet. This one can be used with alternate Pub/Sub endpoints.

feywind:
Looks like this is hitting this PR now: https://github.com/googleapis/nodejs-pubsub/issues/1074

PR #1071, opened 2020-07-27
docs: Add and improve ordering keys examples

kamalaboulhosn:


samples/resumePublish.js
```diff
@@ -0,0 +1,86 @@
...
+ * the Google Cloud Pub/Sub API.
+ *
+ * For more information, see the README.md under /pubsub and the documentation
+ * at https://cloud.google.com/pubsub/docs.
+ */
+
+'use strict';
+
+// sample-metadata:
+//   title: Publish Ordered Message
```
feywind, line 26: Is this the right metadata, or does it need to be updated to be resume-specific?

samples/system-test/topics.test.js
```diff
@@ -188,6 +189,23 @@ describe('topics', () => {
     );
   });
 
+  it('should resume publish', async () => {
+    const topics = require('../resumePublish');
```
feywind, line 193: This is an interesting outlier in the design of the sample tests. Most of the rest of them will just spawn another copy of node to run them, which also tests the command line handling. I think this is probably fine, but I thought I'd note it.

feywind (approved):
I think this looks good overall, sorry for missing it earlier. I think the "merged master into master" update made me glance past it as already merged.

There are a few comments, but I'm good with it as-is too if it all looks right.

samples/resumePublish.js
kamalaboulhosn: Fixed, thanks!

samples/system-test/topics.test.js
kamalaboulhosn: It turns out that there is a test in subscription.test.js, which is where I copied from, that does this. I have changed it to be consistent with the rest of the tests in this file.

PR #1078, opened 2020-08-07
feat: Opentelemetry integration

### LINKED ISSUE #1061: Add OpenTelemetry Support
OpenTelemetry should be added to PubSub to provide OpenTracing and OpenCensus functionality as one API and SDK. It would be helpful to use it to trace a message throughout its life cycle. This is being added to the Python Pub/Sub client and would be useful to have in the node client as well.

https://github.com/googleapis/python-pubsub/pull/149

sethmaxwl:
Adds OpenTelemetry tracing to messages in order to provide more visibility into how messages are behaving behind the scenes. @opentelemetry/api and @opentelemetry/tracing are added as dependencies.

Fixes #1061


src/publisher/index.ts
```diff
@@ -136,7 +140,29 @@ export class Publisher {
    * @param {PublishCallback} [callback] Callback function.
    */
   publishMessage(message: PubsubMessage, callback: PublishCallback): void {
+    // Construct publisher span and set context as message attribute
     const {data, attributes = {}} = message;
+    const spanAttributes = {
```
bcoe, line 145: If we made telemetry configurable, then this logic could be pulled into a conditional block.

Either way I would consider pulling this into a private helper.

src/subscriber.ts
```diff
@@ -445,12 +449,35 @@ export class Subscriber extends EventEmitter {
     for (const data of receivedMessages!) {
       const message = new Message(this, data);
 
+      // Create a new OpenTelemetry span with the publisher span set as the parent
+      const spanValue = message.attributes
```
bcoe, line 453: likewise, could this be pulled into a helper, and start by being conditional.

bcoe (changes requested):
I left a few notes/nits.

My main feedback is that I would start by making the open telemetry feature optional. If you made it configurable with an environment variable, we could start by having select customers enable this environment variable places like GCF/Cloud Run.

I want to be very mindful of customers who are running PubSub out of GCP environments, e.g., other clouds.

src/publisher/index.ts
```diff
@@ -60,11 +62,13 @@ export class Publisher {
   settings!: PublishOptions;
   queue: Queue;
   orderedQueues: Map<string, OrderedQueue>;
+  tracing: OpenTelemetryTracer;
   constructor(topic: Topic, options?: PublishOptions) {
     this.setOptions(options);
     this.topic = topic;
     this.queue = new Queue(this);
     this.orderedQueues = new Map();
+    this.tracing = new OpenTelemetryTracer();
```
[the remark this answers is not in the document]
feywind: Yeah, I'd love for there to be minimal surprises to existing users. Something like this maybe...

```
const pubSub = new PubSub({
  enableTelemetry: true
});
```

feywind:
Thanks for doing this @sethmaxwl and also thanks for the review @bcoe! I agree with what Ben's saying, re: what I wrote above, that we want this to basically be invisible for existing users until they opt in. It sounds like the dependencies are okay, we just want to avoid having it do anything new without an opt-in.

The CI issues look like they're mostly about an ongoing TypeScript version issue. I'm working on a fix for that so we can get several PRs back on track.

samples/package.json
```diff
@@ -14,7 +14,9 @@
     "test": "mocha system-test --timeout 600000"
   },
   "dependencies": {
-    "@google-cloud/pubsub": "^2.3.0"
+    "@google-cloud/pubsub": "^2.3.0",
+    "@opentelemetry/api": "^0.10.2",
```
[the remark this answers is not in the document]
sethmaxwl: The functionality that is already there is not likely to be altered in any significant way. Most of what is still being added to the library is new features or bug fixes.

samples/package.json
```diff
@@ -14,6 +14,8 @@
     "test": "mocha system-test --timeout 600000"
   },
   "dependencies": {
+    "@opentelemetry/api": "^0.10.2",
+    "@opentelemetry/tracing": "^0.10.2"
```
feywind, line 18: Ah @sethmaxwl it looks like this is what the linter is complaining about. There are now two lines in a row with no comma.

package.json
```diff
@@ -30,7 +30,7 @@
     "presystem-test": "npm run compile",
     "system-test": "mocha build/system-test --timeout 600000",
     "samples-test": "cd samples/ && npm link ../ && npm install && npm test && cd ../",
-    "test": "c8 mocha build/test",
+    "test": "c8 mocha build/test --recursive",
```
feywind, line 33: We've got a change pending actually that will obsolete this, but I will deal with it when that gets merged.

samples/opentelemetryTracing.js
```diff
@@ -0,0 +1,97 @@
...
+    };
+
+    // Listens for new messages from the topic
+    pubSubClient.subscription(subscriptionName).on('message', messageHandler);
+    setTimeout(() => {
+      pubSubClient.subscription(subscriptionName).removeAllListeners();
+    }, SUBSCRIBER_TIMEOUT * 1000);
+  }
+
+  publishMessage().then(subscriptionListen());
```
feywind, line 94: This is missing a doc end tag, though it may not matter at this point if nothing in the docsite is expecting it to be here. (We would need to canonicalize the the tag name anyway.)

feywind (approved):
This is looking pretty good, just left a couple more nits...

Let me see if I can get kokoro running the sample tests.

package.json
sethmaxwl: I just added it here to get CI to run the publisher tests too.

samples/opentelemetryTracing.js
sethmaxwl: I want to include a link to this example in the blog post. How should we canonicalize the tag name?

feywind: Let me see if I can find the right person to ask... there were some changes in that area lately.

bcoe (approved):
Thanks for addressing my feedback 👌

PR #1108, opened 2020-09-12
build: test config updates

bcoe:
This makes a couple small tweaks to our test configuration:

1. mocha will now find test files recursively.
2. npm will fail if an attempt is made to install a dependency that is not supported by our engines field.


feywind (approved):
LGTM

PR #1119, opened 2020-10-03
docs: sample audit updates

### LINKED ISSUE #1115: samples: samples audit and cleanup
1. Write a sample for `
pubsub_subscriber_async_pull_custom_attributes` ([docs](https://cloud.google.com/pubsub/docs/pull#pull-messages-with-custom-attributes))
1. Remove the region tag `pubsub_publish` but keep the other region tag `pubsub_quickstart_publish` and keep the sample.
1. Add `pubsub_publish_with_error_handler` ([docs](https://cloud.google.com/pubsub/docs/publisher#publishing_messages)) to show how to handle publish failures. The Java/Python/Go samples are good references. (The existing one shows how to print publish exceptions, which may be enough. If that's the case, simply rename `pubsub_publish` to `pubsub_publish_with_error_handler`.)
1. Rename `pubsub_subscriber_sync_pull_with_lease_management` to `pubsub_subscriber_sync_pull_with_lease`.
1. Fix the incorrect comment about ordering in the ordering key samples.
1. Might also be good to improve the quickstart - https://github.com/googleapis/nodejs-pubsub/issues/1117

### LINKED ISSUE #1117: Better example in the Readme.md
Hello! 
 The current quickstart example is great but it would be so much better if you could see a full workflow
1.  instantiate a client
2. create a topic
3. publish a message
4. create a subscription to that topic
5. receive the message (pull or push, whatever's easier)

I think doing this wouldnt increase the documentation by more than a few lines, however it woudl reduce a lot of confusion and solve developer problems earlier than forcing them to crawl through the whole documentation,

Thank you!

```
Using the client library
// Imports the Google Cloud client library
const {PubSub} = require('@google-cloud/pubsub');

async function quickstart(
  projectId = 'your-project-id', // Your Google Cloud Platform project ID
  topicName = 'my-topic' // Name for the new topic to create
) {
  // Instantiates a client
  const pubsub = new PubSub({projectId});

  // Creates the new topic
  const [topic] = await pubsub.createTopic(topicName);
  console.log(`Topic ${topic.name} created.`);
}
```

feywind:
Fixes: https://github.com/googleapis/nodejs-pubsub/issues/1115
Fixes: https://github.com/googleapis/nodejs-pubsub/issues/1117

Still in progress: `a once-over after getting everything in`


feywind:
@tianzi I gave this its once-over if you want to take a look.

samples/listenForMessages.js
```diff
@@ -28,6 +28,8 @@
 //   usage: node listenForMessages.js <subscription-name> [timeout-in-seconds]
 
 function main(subscriptionName = 'YOUR_SUBSCRIPTION_NAME', timeout = 60) {
+  timeout = Number(timeout);
```
JustinBeckwith, line 31: Is this being added here for the sake of argument validation?  Or do we need to do the coercion here for some reason?  Asking because we don't tend to do argument validation in the sample code, and it's one of those things I think we should probably avoid.

samples/listenForOrderedMessages.js
```diff
@@ -1,124 +0,0 @@
...
-
-const outstandingMessages = {};
-
-async function main(
-  subscriptionName = 'YOUR_SUBSCRIPTION_NAME',
-  timeout = 1000
-) {
-  timeout = Number(timeout);
-
-  // [START pubsub_listen_ordered_messages]
```
JustinBeckwith, line 49: Just to verify - if we search for this on devrel/snippets, we're not breaking anything?

samples/quickstart.js
```diff
@@ -17,24 +17,51 @@
 // sample-metadata:
 //   title: Quickstart
 //   description: A quick introduction to using the Pub/Sub client library.
-//   usage: node quickstart.js <project-id> <topic-name>
+//   usage: node quickstart.js <project-id> <topic-name> <subscription-name>
+
+function exit(code) {
+  // eslint-disable-next-line no-process-exit
+  process.exit(code);
```
JustinBeckwith, line 24: Having a function that calls a single line of code feels weird.  If this is to avoid copying the eslint ignore rule, I'd suggest upleveling it to `samples/.eslintrc.json` as a custom rule.

samples/quickstart.js
```diff
@@ -17,24 +17,51 @@
...
   const [topic] = await pubsub.createTopic(topicName);
   console.log(`Topic ${topic.name} created.`);
+
+  // Creates a subscription on that new topic
+  const [subscription] = await topic.createSubscription(subscriptionName);
+
+  // Receive callbacks for new messages on the subscription
+  subscription.on('message', message => {
+    console.log('Received message:', message.data.toString());
+    exit(0);
```
JustinBeckwith, line 49: Yeah I think just having the `process.exit` here would be way more clear

samples/quickstart.js
```diff
@@ -17,24 +17,51 @@
...
+  });
+
+  // Receive callbacks for errors on the subscription
+  subscription.on('error', error => {
+    console.error('Received error:', error);
+    exit(1);
+  });
+
+  // Send a message to the topic
+  topic.publish(Buffer.from('Test message!'));
```
JustinBeckwith, line 59: I am still sad that the user has to do this :(  Couldn't we check the typeof the arugment, and based on that convert it to a buffer or not?

samples/system-test/topics.test.js
```diff
@@ -83,16 +94,16 @@ describe('topics', () => {
   it('should list topics', async () => {
     const output = execSync(`${commandFor('listAllTopics')}`);
     assert.match(output, /Topics:/);
-    assert.match(output, new RegExp(fullTopicNameOne));
+    assert.match(output, new RegExp(fullTopicNameThree));
```
JustinBeckwith, line 97: For these places where we're not actually using a regex, `assert.include` feels more clear to me

anguillanneuf (approved):
Thanks for the sample update, and cleanup! 
These meet the requirements in the issues I opened.

samples/listenForMessages.js
feywind: Yeah, basically if you pass a timeout on the command line, it will be a string. I honestly try my hardest not to indulge in JavaScript string coercion, but this is the only place it's used. Will that work?

```
    setTimeout(() => {
      subscription.removeListener('message', messageHandler);
      console.log(`${messageCount} message(s) received.`);
    }, timeout * 1000);
```

samples/listenForOrderedMessages.js
feywind: @anguillanneuf ?

samples/quickstart.js
feywind: I feel pretty weird about this too, for a number of reasons. I'm good with making it a linter rule change, I just don't like to presume that stuff :)

samples/quickstart.js
feywind: I like that idea, though I think I'd want to pursue it separate from this change.

samples/listenForOrderedMessages.js
anguillanneuf: Yes temporarily. It's okay because I'm preparing to submit a CL to fix docs right away.

samples/quickstart.js
feywind: https://github.com/googleapis/nodejs-pubsub/issues/1127

samples/listenForMessages.js
feywind: It turns out that "50"*1000 is indeed a number, but it just feels so gross. I was willing to let that go for sample simplicity, but there are some places where I think it will genuinely cause issues, like samples where the value is passed directly into JSON that is sent elsewhere. So I'm back to thinking this normalization is okay, as long as it happens outside the region tags. The samples on the site will still look clean, and the .js files are already kinda gooped with exception handling and such, so I don't feel too bad about that.

PR #1147, opened 2020-11-06
feat: Enable server side flow control by default with the option to turn it off

fayssalmartanigcp:
This change enables sending flow control settings automatically to the server.
If flowControl.maxMessages > 0 or flowControl.maxBytes > 0, flow control will
be enforced at the server side (in addition to the client side).

This behavior is enabled by default and SubscriberOptions.useLegacyFlowControl can be used for users who would
like to opt-out of this feature in case they encounter issues with server side
flow control.

### DIFF
--- a/src/message-stream.ts
+++ b/src/message-stream.ts
@@ -226,8 +226,12 @@ export class MessageStream extends PassThrough {
     const request: StreamingPullRequest = {
       subscription: this._subscriber.name,
       streamAckDeadlineSeconds: this._subscriber.ackDeadline,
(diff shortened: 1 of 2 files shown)


feywind (approved):
Thanks for the contribution! This sort of touches on a conversation we've been having with the team about API surfaces having some backwards compatibility.

fayssalmartanigcp:
Hi, the tests seem to be pending for abnormally long time. Can we check if they are blocked?

Thanks.

PR #1153, opened 2020-11-17
fix: node 14+ changes how multiple destroy() calls work

### LINKED ISSUE #1154: Some old cruft plus Node 14 changes broke stream tests
This is mostly a tracking issue for the fix below. It's currently stopping us from merging any other PRs.

Fixed by: https://github.com/googleapis/nodejs-pubsub/pull/1153

feywind:
Also, this relied on node <14 behaviour, so there were weird conflicts with the standard stream library. There was also some cruft from Node <8 support that needs to go away with the implementation being normalized.

### DIFF
--- a/src/message-stream.ts
+++ b/src/message-stream.ts
@@ -127,7 +127,6 @@ export interface MessageStreamOptions {
  * @param {MessageStreamOptions} [options] The message stream options.
  */
 export class MessageStream extends PassThrough {
-  destroyed: boolean;
   private _keepAliveHandle: NodeJS.Timer;
   private _fillHandle?: NodeJS.Timer;
   private _options: MessageStreamOptions;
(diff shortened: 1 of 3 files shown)


alexander-fenster:
@feywind I hacked a little bit in your branch to make the CI pass here. It indeed fixes the tests, thank you!

src/message-stream.ts
```diff
@@ -156,14 +154,27 @@ export class MessageStream extends PassThrough {
   /**
    * Destroys the stream and any underlying streams.
    *
-   * @param {error?} err An error to emit, if any.
+   * @param {error?} _error An error to emit, if any.
    * @private
    */
-  destroy(err?: Error): void {
+  destroy(_error?: Error | null): void {
```
alexander-fenster, line 160: style nit here and below, why underscores?

alexander-fenster (approved):

src/message-stream.ts
feywind: Oh, this was more or less copied out of the definitions from the Node stream library. I don't think there's a particular reason for it, so I'll move them back to no underscores.

PR #1181, opened 2021-01-08
fix(deps): roll back opentelemetry to 0.12.x for now, as 0.13.x introduces a breaking API change

feywind:
We'll wait for someone with more OT experience to manually update this.
https://github.com/open-telemetry/opentelemetry-js/commit/18b49af9571775c4aaabb7eaa1cb4563ffedb1d9


feywind:
Actually in 0.x, ^ works the same way as ~ in 1.x+ so I'll just close this and not do the 0.13.0 upgrade for now.

PR #1194, opened 2021-01-21
fix(samples): Update synchronousPull to avoid INVALID_ARGUMENT error

psilospore:
I did a copy-pasta and ended up getting this error: `3 INVALID_ARGUMENT: You have not specified an ack ID in the request.`
I have updated the example to check if any messages were received before sending an ack.

I can update any other examples if this checks out.

I know the template asks to make an issue but I thought this might be minor enough to just submit a PR. I can make one if you all would like.


feywind:
@psilospore Thanks, this looks good!

feywind (approved):

PR #1216, opened 2021-02-21
feat: export PublishError

### LINKED ISSUE #1215: Expose PublishError
`PublishError` class is not exported by the package. This blocks developer from pointing using it as a type for method signature in typescript:

What I would like to achieve is:

```js
import { PublishError } from '@google-cloud/pubsub'

export const onError = (error: PublishError): void => {
   ...
}
```

Currently, I can only achieve this by importing from module src, which is not ideal...

```js
import { PublishError } from '@google-cloud/pubsub/build/src/publisher/publish-error'
```

pkuczynski:
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/nodejs-pubsub/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass

Fixes #1215

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -91,7 +91,7 @@ export {
   IAM,
 } from './iam';
 export {Attributes, PublishCallback} from './publisher';
-export {PublishError} from './publisher/publish-error'
+export {PublishError} from './publisher/publish-error';
 export {
   PageOptions,
   GetSnapshotsCallback,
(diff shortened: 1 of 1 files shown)


pkuczynski:
@googlebot I signed it!

feywind (approved):

PR #1225, opened 2021-03-09
docs: add typings, plus new documentation, on the on() methods of Subscriber

feywind:
This adds some (hopefully optional) typings for the subscriber.on() method, as well as some JSDocs about the parameter types. I need to get some double-checking that this isn't going to end up being a breaking change, which I'm a bit nervous about after having to fix the implicit-any in subscription.ts.

```
  private _listen(): void {
    this.on('newListener', event => {
      if (!this.isOpen && event === 'message') {
```

If that ends up being a problem, we can also just comment out the typings for now. They're mostly for IDE support anyway.

I don't think there's an external issue for this, but it's based on b/74085082 (user report).


src/subscription.ts
```diff
@@ -88,6 +88,20 @@ export type SetSubscriptionMetadataResponse = MetadataResponse;
 export type DetachSubscriptionCallback = EmptyCallback;
 export type DetachSubscriptionResponse = EmptyResponse;
 
+// JSDoc won't see these, so this is just to let you get typings
+// in your editor of choice. Callback types are open ended so that
+// not-quite-matching parameters won't cause a breaking change. We
+// may still shift these to required later.
+export declare interface Subscription {
+  on(event: 'message', listener: Function | ((message: Message) => void)): this;
```
JustinBeckwith, line 96: The use of `Function | <blergh>` to me here is strange, because you're providing a nice function type for the user, and then making it more broadly just "Function".  Can we get rid of the `Function |` piece, and always keep stricter bindings?

src/subscription.ts
```diff
@@ -1064,7 +1087,7 @@ export class Subscription extends EventEmitter {
    * @private
    */
   private _listen(): void {
-    this.on('newListener', event => {
+    this.on('newListener', (event: string) => {
```
JustinBeckwith, line 1090: This was my hint above that something was amiss.  I would expect the types to just figure out this is a string.

src/subscription.ts
```diff
@@ -88,6 +88,20 @@ export type SetSubscriptionMetadataResponse = MetadataResponse;
...
+// in your editor of choice. Callback types are open ended so that
+// not-quite-matching parameters won't cause a breaking change. We
+// may still shift these to required later.
+export declare interface Subscription {
+  on(event: 'message', listener: Function | ((message: Message) => void)): this;
+  on(event: 'error', listener: Function | ((error: Error) => void)): this;
+  on(event: 'close', listener: Function): this;
+
+  // Only used internally.
+  on(event: 'newListener', listener: Function): this;
```
bcoe, line 101: you could tighten this up more and declare the function signature:

`(event: string): void`

src/subscription.ts
feywind: Mentioned on chat, but bringing it here for reference: without the `Function` it becomes a compile time breaking change for TypeScript users, because they may be passing in valid (but non-matching) callback handlers. For example, you could have a unit test that just passes in `resolve` to `on()`, and that would work before, but may not work now. My intent here is basically to give IDEs a breadcrumb trail to show the user the types, without breaking any existing code.

src/subscription.ts
feywind: Yeah. Previously, `on()` was basically typed as Ben suggested below, since it's inherited from `EventEmitter`.

src/subscription.ts
feywind: I could combine a few of these, like:

```
on(event: 'close'|'newListener'|etc, listener: Function): this;
```

But in general they need to be broken out, since my intent is to show the available events and what their parameters are.

src/subscription.ts
JustinBeckwith: Ah, in that case - I would still drop `Function` (I think) and move to `(message?: Message) => void` so the parameter is optional.  Would that get us where we want to go?

ettancos: As a user I would expect the types to show whats possible in the api, optional when in fact it is not, would just confuse me. If its possible that on('message') will call my handler without a message parameter than the optional should be there to reflect that, on the other hand if there is no such case, then optional would make me think I have to handle a case that is not possible.
e.g: how it is done for node [process](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/process.d.ts#L309) is very clear
Test use-cases I would not consider for the legit types. Promise resolve is not going to break with a message, but in tests also any stub should work

my 2 cents from the user perspective, hope it helps :)

src/subscription.ts
feywind: I added a catch-all at the bottom that's impossible to match, to give somewhat better error messages. Otherwise the errors were all about `on('removeListener')` with it being at the bottom.

samples/package.json
```diff
@@ -14,9 +14,9 @@
     "test": "mocha system-test --timeout 600000"
   },
   "dependencies": {
+    "@google-cloud/pubsub": "^2.10.0",
     "@opentelemetry/api": "^0.11.0",
-    "@opentelemetry/tracing": "^0.11.0",
-    "@google-cloud/pubsub": "^2.10.0"
+    "@opentelemetry/tracing": "^0.11.0"
```
feywind, line 19: I don't know why npm keeps swapping these around, but I finally just agreed to it :P

JustinBeckwith: npm sometimes likes to put things in alphabetical order for you :)

src/subscription.ts
```diff
@@ -88,6 +89,30 @@ export type SetSubscriptionMetadataResponse = MetadataResponse;
...
 export type DetachSubscriptionResponse = EmptyResponse;
 
+// JSDoc won't see these, so this is just to let you get typings
+// in your editor of choice. Callback types are open ended so that
+// not-quite-matching parameters won't cause a breaking change. We
+// may still shift these to required later.
+export declare interface Subscription {
+  on(
+    event: 'message',
+    listener: (() => void) | ((message: Message) => void)
```
JustinBeckwith, line 99: You could simplify this by collapsing to:  `listener: (message?: Message) => void`

JustinBeckwith: Here and below

src/subscription.ts
```diff
@@ -88,6 +89,30 @@ export type SetSubscriptionMetadataResponse = MetadataResponse;
...
+  ): this;
+  on(event: 'close', listener: () => void): this;
+
+  // Only used internally.
+  on(event: 'newListener', listener: Function): this;
+  on(event: 'removeListener', listener: Function): this;
+
+  // Catch-all. If you get an error about this line, it means you're
+  // using an unsupported event type or listener type.
+  on(event: string, listener: void): this;
```
JustinBeckwith, line 113: I don't fully understand why this is needed.  Are you saying that without it, using a non-enumerated event name doesn't cause a compile time error?

src/subscription.ts
JustinBeckwith: Still don't think we want to require the typing here.

JustinBeckwith (commented):
Happy to hop on a quick chat if you want to tighten the loop on communication for this one, sorry for the delay.

src/subscription.ts
feywind: Yeah, I did this in response to @ettancos 's comment. It's sort of misleading to say that `Message` is optional here, because it's not. It feels like a weird hack to get around a TS limitation.

At this point I think I'm prepared to just comment out the typings part for now, and re-add them on the next major. Getting the JSDoc in place is more immediately useful, and I think all of us are nervous about breaking existing clients with typings (for various reasons).

src/subscription.ts
feywind: TS produces unhelpful error messages otherwise, if you use a non-enumerated event, or use a bad type for the callback on an enumerated one. It'll say that your call to `on()` doesn't match the definition for whatever `on('foo')` is declared last.

JustinBeckwith (approved):

PR #1234, opened 2021-03-17
fix: added support for Opentelemetry 0.18

### LINKED ISSUE #1233: Stops working when project is using opentelemetry-js 0.18+ packages

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

1) Is this a client library issue or a product issue?
Yes, this is a client library issue

2) Did someone already solve this?
There renovatebot made a PR to solve this problem

3) Do you have a support contract?
No

#### Environment details

  - OS: macOS 11.2.3
  - Node.js version: 14.16.0
  - npm version: 6.x
  - `@google-cloud/pubsub` version: 2.10.0

#### Steps to reproduce

  1. Create new directory
  2. `npm init -y`
  3. Install the dependencies `npm install @google-cloud/pubsub @opentelemetry/core @opentelemetry/api`
  4. `npm ls @opentelemetry/tracing`
  5.  Notice that the installed version is 0.12 instead of the 0.18 version meets the Opentelemetry 1.0.0 tracing spec

Making sure to follow these steps will guarantee the quickest resolution possible.

Thanks!

weyert:
Introduces support for Opentelemetry 0.18 API which implements the
Tracing 1.0 specification.

- [X] Ensure the tests and linter pass

Fixes #1233  🦕


weyert:
@googlebot I consent

tapico-weyert:
@googlebot I consent

src/opentelemetry-tracing.ts
```diff
@@ -31,13 +38,21 @@ export class OpenTelemetryTracer {
...
     const tracerProvider: Tracer = trace.getTracer('default') as Tracer;
-    return tracerProvider.startSpan(spanName, {
-      parent: parent,
-      attributes: attributes,
-    });
+
+    let spanContext = undefined;
+    if (parent) {
+      spanContext = setSpanContext(context.active(), parent);
+    }
```
aabmass, line 49: I'd recommend just setting parent in context immediately when you extract it from the message. Then you don't need to pass it here, it should be implicit.

Looks like this whole class could just be a function as well, if you have time for a little refactor 😄

weyert: I have applied your feedback, let me know if this is what you had in mind :)

weyert: There is one test which even in `master` is incorrect but I can't seem to make it work either so not sure what to do. It's related to the subscriber's test named: 'exports a span once it is created' which asserts `assert.ok(spans)` which is `true` when `exporter.getFinishedSpans()` an empty array and its not explicitly checking if at least one span is returned.

If you or anyone else knows how to fix this, that would be great! I am not seeing it.

package.json
```diff
@@ -66,6 +65,7 @@
     "p-defer": "^3.0.0"
   },
   "devDependencies": {
+    "@opentelemetry/tracing": "^0.18.0",
```
aabmass, line 68: 👍

test/publisher/index.ts
```diff
@@ -223,7 +192,15 @@ describe('Publisher', () => {
       opentelemetry.trace.setGlobalTracerProvider(provider);
 
       tracingPublisher.publish(buffer);
-      assert.ok(exporter.getFinishedSpans());
+      const spans = exporter.getFinishedSpans();
```
aabmass, line 195: Nice this looks better

test/subscriber.ts
```diff
@@ -648,33 +648,35 @@ describe('Subscriber', () => {
...
+      assert.strictEqual(tracingSubscriber['_useOpentelemetry'], false);
     });
 
     it('should instantiate a tracer when tracing is enabled through constructor', () => {
       tracingSubscriber = new Subscriber(subscription, enableTracing);
-      assert.ok(tracingSubscriber['_tracing']);
+      assert.ok(tracingSubscriber['_useOpentelemetry']);
     });
 
     it('should instantiate a tracer when tracing is enabled through setOptions', () => {
```
aabmass, line 664: I think this "instantiate a tracer" needs to be rephrased with this change. It will always instantiate a tracer, but it will be a noop

package.json
weyert: Yeah, think that's causing most of the problems I am experiencing

aabmass (commented):
Question to the maintainers here, can we remove `enableOpenTelemetryTracing` option altogether?

Enabling instrumentation in OpenTelemetry should be done by
1. User adding a dependency on the SDK (instead of just the API) `@opentelemetry/tracing` (which is why @weyert has moved it out of `dependencies`)
2. User configuring the global tracer provider to the SDK `TracerProvider` as is done here: https://github.com/googleapis/nodejs-pubsub/blob/883828800c94f7ea21c8306d272b70b4576c664c/samples/openTelemetryTracing.js#L66
    An alternative is the user can pass a tracer provider instance as an option, otherwise we get the global one.

If the user doesn't take those steps, the `@opentelemetry/api` package will give no-op implementations that don't do anything. It should be implicit

test/publisher/index.ts
weyert: yes, the problem is that it wasn't checking if it had a span so it always was successful (even now in master) if I change this to check the length >= 1 it will fail. 

Only I am not well versed in this library or its tests to see were the issue is. It does like that _constructSpan always exits early due the first if-statement.

test/subscriber.ts
weyert: originally i was thinking to just remove these tests 🙈

feywind:
Thanks for all the code and comments, I appreciate it... I haven't really known what to do with the OT support since the APIs changed. It was an outside contribution, so none of us know that API too well. So to answer @aabmass 's comment, I'm open to guidance on where to take it over time.

Busy day today, I'll look at this in a bit! :)

src/opentelemetry-tracing.ts
```diff
@@ -13,31 +13,35 @@
...
-  ): Span {
-    const tracerProvider: Tracer = trace.getTracer('default') as Tracer;
-    return tracerProvider.startSpan(spanName, {
-      parent: parent,
+export function createSpan(
+  spanName: string,
+  attributes?: SpanAttributes,
+  parent?: SpanContext
+): Span {
+  const tracerProvider: Tracer = trace.getTracer('default') as Tracer;
```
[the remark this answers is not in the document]
weyert: Removed that

weyert:
> Thanks for all the code and comments, I appreciate it... I haven't really known what to do with the OT support since the APIs changed. It was an outside contribution, so none of us know that API too well. So to answer @aabmass 's comment, I'm open to guidance on where to take it over time.
> 
> Busy day today, I'll look at this in a bit! :)

Yes, the API shouldn't really happen anymore as the API is 1.0.0 now and it shouldn't change much anymore and I think only backward compatible changes are allowed and 1.0.0 will be supported for a few years. The `opentelemetry-js` team is planning to release 1.0.0 rc of the SDK and API soon but not much has changed between 0.18 and 1.0.0-rc. I think migrating to 0.18 is a good start in the right direction.

Currently, I have upgraded to 0.18 and tried to remove the dependency to the `tracing` SDK package which isn't met to be used by libraries. Only during my changes I am have found a broken test which even occurs in the `master`-branch when improving the `assert.ok` to actually check that a span is being created. Until now it looks it only checks it was a thruthly value

test/publisher/index.ts
aabmass: So its still not creating a span?

feywind: Yeah, I tried it a minute ago and it looks like it's still not getting any spans. I thought maybe it's because you have to add a tracer, but that doesn't seem to have changed it. I don't know much about OT either :)

Slightly unrelated: `assert.equal` shows up as deprecated for me, and it suggests `assert.strictEqual`.

aabmass: I'll take a look

test/subscriber.ts
feywind: I guess we could verify that it's a no-op tracer, but if that's the way OT works now, I'm also okay with just removing those tests.

test/publisher/index.ts
aabmass: @weyert I'll open a PR to your fork. I think my original suggestions were too lofty, I want to sync up with the pubsub folks before changing things too much.

test/subscriber.ts
weyert: Yeah, that sounds reasonable. The unit test is using a in-memory trace provider / exporter to determine if the spans get created or not. I am just struggling a bit with mocha and how things are mocked. I can have a stab at it again next week. 

Would you be able with me just removing the whole opt-in settings flag `enableOpentelemetryTracing`?

aabmass:
@weyert the test failure looks to be related to globals and things running before this test. If I only run that case with `npm run test -- -g 'export created spans'` it passes. Might be related to the change I requested for the module level tracer.

I think depending when @feywind wants to make more changes, we should change things a bit. If we alter the configuration here and remove `enableOpentelemetryTracing`, it is a breaking change and would require a major release.

So for now @weyert, lets maybe leave the config in and just upgrade the library to 0.18?

weyert:
> @weyert the test failure looks to be related to globals and things running before this test. If I only run that case with `npm run test -- -g 'export created spans'` it passes. Might be related to the change I requested for the module level tracer.

Thanks for trying this out! I didn't think about trying this out! 
 
> I think depending when @feywind wants to make more changes, we should change things a bit. If we alter the configuration here and remove `enableOpentelemetryTracing`, it is a breaking change and would require a major release.
> 
> So for now @weyert, lets maybe leave the config in and just upgrade the library to 0.18?

Sounds good, so the change to `startSpan` and revert back to using the class?

Sorry, I haven't had a change yet to look into it today.

aabmass:
> Sounds good, so the change to `startSpan` and revert back to using the class?

Hmm you mentioned it was broken before this PR too so maybe that won't fix it :/ but its worth a try

feywind:
Apologies for pushing a change in there, but hopefully this will get it going. I still don't know what I'm doing with OpenTelemetry :D but it seemed like some state was getting set in the earlier test.

weyert:
> Apologies for pushing a change in there, but hopefully this will get it going. I still don't know what I'm doing with OpenTelemetry :D but it seemed like some state was getting set in the earlier test.

Awesome, I hope to have time to look into it tomorrow. Ended up spending most of today trying to find out what was causing this error when using this library in a project. `UnhandledPromiseRejectionWarning: Error: 13 INTERNAL: A service error has occurred. Please retry your request. If the error persists, please report it. [code=e8c0]`

feywind:
Ah. Yeah, I have been seeing more of those in system tests lately, and it's generally something between grpc and the service (or the service itself).

src/opentelemetry-tracing.ts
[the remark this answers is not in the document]
feywind: I just realized my change undid this suggestion, but that's actually what made the test work again. Maybe something else is needed there still?

test/subscriber.ts
feywind: I was thinking about this too... the parts that call OT aren't really a problem, but currently it's adding a message attribute to things sent through Pub/Sub. I haven't dug into it enough to see if that gets removed again later or not, but it might increase a user's bandwidth or storage unintentionally. Maybe it can just be repurposed to enable those attributes for correlating on the server side.

weyert:
> Ah. Yeah, I have been seeing more of those in system tests lately, and it's generally something between grpc and the service (or the service itself).

Off-topic, but seems to happen a lot when trying to call `UpdateSubscription` GRPC method e.g. to set the dead letter policy and trying in quick succession in different GKE pods. But I am drifting away from the subject 😅

feywind:
Putting this link over here as a breadcrumb trail in case anyone searches for that internal error and wonders where the discussion went :) https://github.com/googleapis/nodejs-pubsub/issues/1219

weyert:
Just to  let everyone that I am looking into this issue today :) Work things got into the way, like it always goes ;)

feywind:
Sorry, I should pass on the sample test failure:

```
  36 passing (6m)
  1 failing

  1) openTelemetry
       should run the openTelemetryTracing sample:
     AssertionError: expected 'Message 2190434663794456 published.\nMessage 2190434663794456 received.\n' to match /traceId/
      at Context.<anonymous> (system-test/openTelemetryTracing.test.js:45:12)
      at processImmediate (internal/timers.js:456:21)
```

It looks like the traceId attribute isn't being added to these messages.

weyert:
Finally got it working and now it doesn't work when it runs with all the other tests 😭

weyert:
@feywind I have been able to make the tests past but now its complaining about header check. Any idea what's wrong?

weyert:
CleanShot 2021-03-31 at 05 16 18@2x

feywind:
> Sorted it, needed to be Google LLC. Not sure why it's not complaining about the file I copied it from 🤷‍♀️

Yeah, the short version is that because we're managing like... 200 repos 😱 ...there's a lot of picky automation in place, and it's not always good at noticing new vs old when checking changed files.

test/publisher/index.ts
```diff
@@ -180,56 +184,57 @@ describe('Publisher', () => {
...
+        'span kind should be PRODUCER'
+      );
+      assert.ok(spans);
     });
   });
 
   describe('publishMessage', () => {
     const data = Buffer.from('hello, world!');
-    const spy = sandbox.spy();
+    // const spy = sandbox.spy();
```
feywind, line 237: Double checking - should this line go away? I think the answer is yes, but I want to make sure.

test/subscriber.ts
```diff
@@ -708,20 +714,56 @@ describe('Subscriber', () => {
...
+      );
+      assert.strictEqual(
+        firstSpan.attributes[MessagingAttribute.MESSAGING_DESTINATION],
+        subscriber.name,
+        'span messaging destination should match'
+      );
+      assert.strictEqual(
+        firstSpan.attributes[MessagingAttribute.MESSAGING_DESTINATION_KIND],
+        'topic'
+      );
```
feywind, line 762: All this detail and matching testing is much appreciated!

test/publisher/index.ts
weyert: Yeah, there is now a `spy` defined at the top of the unit test block. Only way I was able to get things work and after four days I started to give up 🙈

test/subscriber.ts
weyert: Thank you! If you know how to get the project id in the `Publisher` that would be awesome to add.

test/tracing.ts
```diff
@@ -0,0 +1,39 @@
...
+} from '@opentelemetry/tracing';
+
+/**
+ * This file is used to initialise a global tracing provider and span exporter
+ * for our tests used in our project. This was the only way I was able to get
+ * the tracing tests to work.
+ *
+ * Now before each test related or touches Opentelemetry
+ * we are resetting the exporter defined below to ensure there are now spans
+ * from previous tests still in memory.
```
feywind, line 30: Comment nitpick: did you mean 'there are no spans'? I stared at that for a while, because I think the intent is that any previous tracers/spans/etc are removed from previous tests. :)

test/subscriber.ts
feywind: It could probably be piped through, but I guess I wouldn't worry about it too much? Getting everything working again seems like a good short term goal.

weyert: Yeah, I will remove it for now :)

weyert: Removed the NET_PEER_NAME attribute for now rest is still there :)

test/tracing.ts
weyert: @feywind  I have updated the comment :)

weyert:
@feywind Think it's all good now :D

feywind (dismissed):
Yay, this is looking good. I'll try to get this merged today. It might be next week for a release, though. I have superstitions about releases before weekends, and double superstitions about releases before long weekends 😹

feywind:
I think this is Kokoro being grumpy:

```
src/index.ts:1:22 - error TS2307: Cannot find module '@google-cloud/pubsub' or its corresponding type declarations.

1 import {PubSub} from '@google-cloud/pubsub';
                       ~~~~~~~~~~~~~~~~~~~~~~
```

I'll retry it, and if it's still not working, I'll try again next week.

weyert:
> Yay, this is looking good. I'll try to get this merged today. It might be next week for a release, though. I have superstitions about releases before weekends, and double superstitions about releases before long weekends 😹

Awesome! Thank you :)) Good start of the long weekend :>

feywind:
Hmm, looks like there might actually be a sample-test issue:

```
  openTelemetry
/workspace/samples/openTelemetryTracing.js:66
  opentelemetry.diag.setLogger(
                     ^

TypeError: Cannot read property 'setLogger' of undefined
    at main (/workspace/samples/openTelemetryTracing.js:66:22)
    at Object.<anonymous> (/workspace/samples/openTelemetryTracing.js:135:1)
    at Module._compile (internal/modules/cjs/loader.js:1137:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1157:10)
    at Module.load (internal/modules/cjs/loader.js:985:32)
    at Function.Module._load (internal/modules/cjs/loader.js:878:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
    at internal/main/run_main_module.js:17:47
    1) should run the openTelemetryTracing sample
```

weyert:
> Hmm, looks like there might actually be a sample-test issue:
> 
> ```
>   openTelemetry
> /workspace/samples/openTelemetryTracing.js:66
>   opentelemetry.diag.setLogger(
>                      ^
> 
> TypeError: Cannot read property 'setLogger' of undefined
>     at main (/workspace/samples/openTelemetryTracing.js:66:22)
>     at Object.<anonymous> (/workspace/samples/openTelemetryTracing.js:135:1)
>     at Module._compile (internal/modules/cjs/loader.js:1137:30)
>     at Object.Module._extensions..js (internal/modules/cjs/loader.js:1157:10)
>     at Module.load (internal/modules/cjs/loader.js:985:32)
>     at Function.Module._load (internal/modules/cjs/loader.js:878:14)
>     at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
>     at internal/main/run_main_module.js:17:47
>     1) should run the openTelemetryTracing sample
> ```

How do you test these?

weyert:
Ah found the `samples-test`, yeah the `package.json` in that directory is having an old version of opentelemetry defined pushed a change for that.

src/publisher/index.ts
```diff
@@ -264,9 +256,23 @@ export class Publisher {
...
+      [MessagingAttribute.MESSAGING_TEMP_DESTINATION]: false,
+      [MessagingAttribute.MESSAGING_SYSTEM]: 'pubsub',
+      [MessagingAttribute.MESSAGING_OPERATION]: '', // operation expected to be empty
+      [MessagingAttribute.MESSAGING_DESTINATION]: this.topic.name,
+      [MessagingAttribute.MESSAGING_DESTINATION_KIND]: 'topic',
+      [MessagingAttribute.MESSAGING_MESSAGE_ID]: message.messageId,
+      [MessagingAttribute.MESSAGING_PROTOCOL]: 'pubsub',
+      [MessagingAttribute.MESSAGING_MESSAGE_PAYLOAD_COMPRESSED_SIZE_BYTES]: 0,
+      [MessagingAttribute.MESSAGING_MESSAGE_PAYLOAD_SIZE_BYTES]:
+        message.data?.length,
```
aabmass, line 269: nit given this check above https://github.com/googleapis/nodejs-pubsub/blob/78f31e4aa637a6ca755627ad3c54cbac59040e39/src/publisher/index.ts#L157

i was a bit confused otherwise if this was just string length 
```suggestion
        (message.data as Buffer).length,
```

src/subscriber.ts
```diff
@@ -447,24 +448,50 @@ export class Subscriber extends EventEmitter {
...
       deliveryAttempt: message.deliveryAttempt,
+      //
+      // based on https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#topic-with-multiple-consumers
+      [MessagingAttribute.MESSAGING_SYSTEM]: 'pubsub',
+      [MessagingAttribute.MESSAGING_OPERATION]: 'process',
+      [MessagingAttribute.MESSAGING_DESTINATION]: this.name,
+      [MessagingAttribute.MESSAGING_DESTINATION_KIND]: 'topic',
+      [MessagingAttribute.MESSAGING_MESSAGE_ID]: message.id,
+      [MessagingAttribute.MESSAGING_PROTOCOL]: 'pubsub',
+      [MessagingAttribute.MESSAGING_MESSAGE_PAYLOAD_COMPRESSED_SIZE_BYTES]: 0,
```
aabmass, line 473: I believe this should be omitted

src/publisher/index.ts
```diff
@@ -264,9 +256,23 @@ export class Publisher {
...
       data: message.data,
+      // Add Opentelemetry semantic convention attributes to the span
+      [MessagingAttribute.MESSAGING_TEMP_DESTINATION]: false,
+      [MessagingAttribute.MESSAGING_SYSTEM]: 'pubsub',
+      [MessagingAttribute.MESSAGING_OPERATION]: '', // operation expected to be empty
+      [MessagingAttribute.MESSAGING_DESTINATION]: this.topic.name,
+      [MessagingAttribute.MESSAGING_DESTINATION_KIND]: 'topic',
+      [MessagingAttribute.MESSAGING_MESSAGE_ID]: message.messageId,
+      [MessagingAttribute.MESSAGING_PROTOCOL]: 'pubsub',
+      [MessagingAttribute.MESSAGING_MESSAGE_PAYLOAD_COMPRESSED_SIZE_BYTES]: 0,
```
aabmass, line 267: I believe this should be omitted

src/subscriber.ts
```diff
@@ -447,24 +448,50 @@ export class Subscriber extends EventEmitter {
...
+      [MessagingAttribute.MESSAGING_MESSAGE_PAYLOAD_SIZE_BYTES]: message.length,
+      // Not in Opentelemetry semantic convention but mimics naming
+      'messaging.consumer.published_at':
+        typeof message.publishTime.getTime !== 'undefined'
+          ? message.publishTime.getTime()
+          : undefined,
```
aabmass, line 479: nit isn't this equivalent?
```suggestion
        message.publishTime?.getTime()
```

src/subscriber.ts
```diff
@@ -447,24 +448,50 @@ export class Subscriber extends EventEmitter {
...
+      'messaging.consumer.received_at': message.received,
+      'messaging.consumer.acknowlege_id': message.ackId,
+      'messaging.consumer.delivery_attempt': message.deliveryAttempt,
     };
+
     // Subscriber spans should always have a publisher span as a parent.
     // Return undefined if no parent is provided
+    const spanName = `${this.name} receive`;
```
aabmass, line 487: The operation name here must match `messaging.operation` attribute https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#user-content-span-name:~:text=The%20values%20allowed%20for%20%3Coperation%20name%3E,defined%20for%20message%20consumer%20spans%20below.

src/publisher/index.ts
```diff
@@ -264,9 +256,23 @@ export class Publisher {
   constructSpan(message: PubsubMessage): Span | undefined {
     const spanAttributes = {
       data: message.data,
```
aabmass, line 258: @feywind I just noticed this was already here but probably shouldn't be written to cloud trace?

test/subscriber.ts
aabmass: Good point, we probably shouldn't be doing that unless its a real trace AND opted into. You can check if its a real vs no-op trace with [`isSpanContextValid()`](https://github.com/open-telemetry/opentelemetry-js-api/blob/92c170009ba36041ec0adba7804ce0a06d1b0330/src/api/trace.ts#L78)

protos/protos.d.ts
```diff
@@ -6210,6 +6210,9 @@ export namespace google {
 
             /** Http rules */
             rules?: (google.api.IHttpRule[]|null);
+
+            /** Http fullyDecodeReservedExpansion */
+            fullyDecodeReservedExpansion?: (boolean|null);
```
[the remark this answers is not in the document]
weyert: Sorry, what do you mean? Looks like it's up to date with master?

src/opentelemetry-tracing.ts
```diff
@@ -13,31 +13,49 @@
...
+  SpanAttributes,
+  SpanContext,
+  Span,
+  context,
+  trace,
+  setSpanContext,
+  SpanKind,
+} from '@opentelemetry/api';
+
+import PKG = require('../package.json');
```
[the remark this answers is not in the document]
weyert: I have updated it but looking forward hearing from @feywind

src/publisher/index.ts
```diff
@@ -264,9 +256,23 @@ export class Publisher {
...
+      // Add Opentelemetry semantic convention attributes to the span
```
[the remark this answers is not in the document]
weyert: Added

src/publisher/index.ts
```diff
@@ -264,9 +256,23 @@ export class Publisher {
...
+      [MessagingAttribute.MESSAGING_TEMP_DESTINATION]: false,
+      [MessagingAttribute.MESSAGING_SYSTEM]: 'pubsub',
+      [MessagingAttribute.MESSAGING_OPERATION]: '', // operation expected to be empty
```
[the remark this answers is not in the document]
weyert: I roughly based it on this example: https://github.com/open-telemetry/opentelemetry-specification/blob/v1.1.0/specification/trace/semantic_conventions/messaging.md#topic-with-multiple-consumers

Do you think this is the wrong one?

src/publisher/index.ts
weyert: Removed it

src/subscriber.ts
weyert: Removed it

src/subscriber.ts
weyert: Updated it so its now `message.publishTime?.getTime?.()`

src/publisher/index.ts
weyert: Awaiting feedback from @feywind. Personally, I think we should remove it as it could potentially unexpectedly leak unwanted PPI information to the service the traces are exported too

src/subscriber.ts
weyert: Updated

samples/openTelemetryTracing.js
```diff
@@ -89,17 +93,32 @@ function main(
     const messageHandler = message => {
       console.log(`Message ${message.id} received.`);
       message.ack();
-      process.exit(0);
+
+      // Ensure that all spans got flushed by the exporter
+      console.log('Cleaning up Opentelemetry exporter...');
+      exporter.shutdown().then(() => {
```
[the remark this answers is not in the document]
weyert: Yeah we could, I thought this was clear enough. Don't want to change too much. I want to avoid failing test 😅

protos/protos.d.ts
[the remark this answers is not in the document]
aabmass: Did you generate these changes in this PR? I'm just confused where it came from

samples/openTelemetryTracing.js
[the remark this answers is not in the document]
aabmass: no worries, feel free to ignore

src/publisher/index.ts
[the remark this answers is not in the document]
aabmass: Feel free to resolve my comments when done (I can't)

src/publisher/index.ts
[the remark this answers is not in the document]
aabmass: Ya it looks like a mistake in spec considering this line:

> The values allowed for <operation name> are defined in the section Operation names below. If the format above is used, the operation name MUST match the messaging.operation attribute defined for message consumer spans below.

protos/protos.d.ts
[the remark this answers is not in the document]
feywind: I think that was probably me, merging in from main. The proto updates end up being a bit messy right now, and the current guidance is to just commit the proto changes.

src/opentelemetry-tracing.ts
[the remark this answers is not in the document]
feywind: The tl;dr is that I'd use `import` as long as it works.

Here is some info on that: https://stackoverflow.com/questions/52534910/difference-between-import-x-requirex-and-const-x-requirex-in-typ

It more or less matches my guess, which is that `import` pulls in types, while `const` doesn't. Both have the same effect at runtime.

Some of the code in this package is really old, and the `import` syntax might not have existed back then, so that's probably why the other module has it.

src/publisher/index.ts
feywind: That does seem a little bit more clear, but I could see it either way. If you did want to make that change, might want something like:

`(message.data as Buffer?)?.length`

...which is looking a bit messy, so maybe the existing one ends up being nicer after all.

src/publisher/index.ts
feywind: It also might end up being a lot of wasted data, if the messages are on the bigger side, so yeah, might be good to take that part out.

src/opentelemetry-tracing.ts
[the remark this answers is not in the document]
weyert: Do you want me to change it?

protos/protos.d.ts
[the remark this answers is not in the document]
weyert: All good then?

src/publisher/index.ts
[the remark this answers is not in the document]
weyert: Okay, happy to change :)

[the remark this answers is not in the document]
weyert: Updated

src/publisher/index.ts
```diff
@@ -264,9 +256,23 @@ export class Publisher {
...
+      [MessagingAttribute.MESSAGING_DESTINATION]: this.topic.name,
+      [MessagingAttribute.MESSAGING_DESTINATION_KIND]: 'topic',
+      [MessagingAttribute.MESSAGING_MESSAGE_ID]: message.messageId,
+      [MessagingAttribute.MESSAGING_PROTOCOL]: 'pubsub',
+      [MessagingAttribute.MESSAGING_MESSAGE_PAYLOAD_COMPRESSED_SIZE_BYTES]: 0,
+      [MessagingAttribute.MESSAGING_MESSAGE_PAYLOAD_SIZE_BYTES]:
+        message.data?.length,
+      'messaging.consumer.published_at': message.publishTime,
```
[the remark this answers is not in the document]
weyert: Yeah, guess it can be removed. I thought it was useful to have it on the span as its part of the message. Removed it

src/publisher/index.ts
```diff
@@ -264,9 +256,23 @@ export class Publisher {
...
+      'messaging.consumer.ordering_key': message.orderingKey,
```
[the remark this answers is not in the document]
weyert:
```
      'messaging.consumer.published_at': message.publishTime,
```
Removed this one

Renamed this one to `pubsub` as suggested:
```
       'messaging.consumer.ordering_key': message.orderingKey,
```

src/publisher/index.ts
```diff
@@ -264,9 +256,23 @@ export class Publisher {
...
     } as Attributes;
-    const span: Span | undefined = this.tracing
-      ? this.tracing.createSpan(`${this.topic.name} publisher`, spanAttributes)
+
+    const span: Span | undefined = this.settings.enableOpenTelemetryTracing
+      ? createSpan(`${this.topic.name} send`, SpanKind.PRODUCER, spanAttributes)
       : undefined;
     if (span) {
```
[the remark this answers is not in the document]
weyert: Updated it

src/publisher/index.ts
weyert: I accepted Aaron suggestion :)

src/opentelemetry-tracing.ts
weyert: This test has been fixed and passing

src/opentelemetry-tracing.ts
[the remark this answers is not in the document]
weyert: This should be resolved now :)

src/publisher/index.ts
weyert: Removed it with a BREAKING CHANGE marker in the commit

src/publisher/index.ts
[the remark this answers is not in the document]
weyert: Removed it

test/subscriber.ts
weyert: Added it

weyert:
@aabmass @feywind I have submitted a bunch of changes. Beside the proto comment I think all the feedback on the PR has been resolved.

weyert:
@feywind @aabmass  Let me know if you are happy with this PR or whether you like more changes

src/publisher/index.ts
```diff
@@ -254,26 +254,31 @@ export class Publisher {
...
+    }
+
     const spanAttributes = {
       data: message.data,
       // Add Opentelemetry semantic convention attributes to the span, based on:
       // https://github.com/open-telemetry/opentelemetry-specification/blob/v1.1.0/specification/trace/semantic_conventions/messaging.md
       [MessagingAttribute.MESSAGING_TEMP_DESTINATION]: false,
       [MessagingAttribute.MESSAGING_SYSTEM]: 'pubsub',
-      [MessagingAttribute.MESSAGING_OPERATION]: '', // operation expected to be empty
+      [MessagingAttribute.MESSAGING_OPERATION]: 'send', // operation expected to be empty
```
feywind, line 266: Should that comment go away?

feywind:
@weyert It looks like things are more or less testing okay, I'll give you another pass to look at that comment and my problematic merge. :S But I think it's looking pretty good.

We had some discussions today about trying to normalize what we send out in OT across different languages, no conclusions there yet, but I think this is fine to start with!

src/publisher/index.ts
weyert: Ah yes, I can remove that

weyert:
> @weyert It looks like things are more or less testing okay, I'll give you another pass to look at that comment and my problematic merge. :S But I think it's looking pretty good.
> 
> We had some discussions today about trying to normalize what we send out in OT across different languages, no conclusions there yet, but I think this is fine to start with!

Awesome! I have removed the comment so that should be all good. I will also solve linting issue

feywind (approved):

src/publisher/index.ts
```diff
@@ -262,13 +254,33 @@ export class Publisher {
...
+      [MessagingAttribute.MESSAGING_MESSAGE_PAYLOAD_SIZE_BYTES]:
+        message.data?.length,
+      'messaging.pubsub.ordering_key': message.orderingKey,
     } as Attributes;
-    const span: Span | undefined = this.tracing
-      ? this.tracing.createSpan(`${this.topic.name} publisher`, spanAttributes)
-      : undefined;
-    if (span) {
+
+    const span: Span | undefined = createSpan(
```
aabmass, line 276:
```suggestion
    const span: Span = createSpan(
```

aabmass (approved):
Thanks for being patient with all of the review comments, Weyert 😄 LGTM!

weyert:
> Thanks for being patient with all of the review comments, Weyert 😄 LGTM!

Thank you! Glad you are happy with it. I hope we can finally merge it! :)
Do you think we want to upgrade the `opentelemetry` dependencies to 0.18.2 or leave it for now?

weyert:
@feywind downgrading `@sinonjs/fake-timers`-package did the trick build appears to be passing now :)

aabmass:
> Do you think we want to upgrade the `opentelemetry` dependencies to 0.18.2 or leave it for now?

If they are compatible for the purposes of this library, I think it's fine to leave it at 0.18.2 since we are using caret dependency

feywind:
The kokoro tests are failing because of this error:

```
src/index.ts:1:22 - error TS2307: Cannot find module '@google-cloud/pubsub' or its corresponding type declarations.

1 import {PubSub} from '@google-cloud/pubsub';
                       ~~~~~~~~~~~~~~~~~~~~~~
```

I have no idea what's going on there yet.

What was the deal with the sinon fake timers module? It seems like we've been on sinon 10 for a bit and that's compiling in other branches. I'm sort of loathe to pin deep dependencies since they sometimes end up forgotten later :)

weyert:
> The kokoro tests are failing because of this error:
> 
> ```
> src/index.ts:1:22 - error TS2307: Cannot find module '@google-cloud/pubsub' or its corresponding type declarations.
> 
> 1 import {PubSub} from '@google-cloud/pubsub';
>                        ~~~~~~~~~~~~~~~~~~~~~~
> ```
> 
> I have no idea what's going on there yet.
> 
> What was the deal with the sinon fake timers module? It seems like we've been on sinon 10 for a bit and that's compiling in other branches. I'm sort of loathe to pin deep dependencies since they sometimes end up forgotten later :)

They changed the type definitions in the latest version

weyert:
Okay, think I found the problem. As part of the system test its running `npm pack` and locally this gives me the following output:

```
npm notice 
npm notice 📦  @google-cloud/pubsub@2.10.0
npm notice === Tarball Contents === 
npm notice 11.4kB LICENSE           
npm notice 3.6kB  build/package.json
npm notice 3.2kB  package.json      
npm notice 57.5kB CHANGELOG.md      
npm notice 20.2kB README.md         
npm notice === Tarball Details === 
npm notice name:          @google-cloud/pubsub                    
npm notice version:       2.10.0                                  
npm notice filename:      google-cloud-pubsub-2.10.0.tgz          
npm notice package size:  22.0 kB                                 
npm notice unpacked size: 95.8 kB                                 
npm notice shasum:        3c77e1e46cfcbccb3cbd2de732d81a6fc5ff7823
npm notice integrity:     sha512-3H4qqXZWOSogu[...]6ERB1uWfewv2g==
npm notice total files:   5                                       
npm notice 
google-cloud-pubsub-2.10.0.tgz
```

As you can see it's not bundling the `build`-directories anymore so it's correct it can't find the type definitions are they aren't included. The cause of it not being bundled anymore appears to be that when the `compile`-task is run a preparation task for running `pack` it includes a `package.json` in the `build`-directory and stops it from including these directories. If I first remove `./build/package.json` it appears to work. 

The compile step seems to copy it over to the `build`-task I think this is because I changed the way how we are importing `package.json` from using `require` to `import`.

weyert:
Hmm, the `ci/kokoro`-jobs are not running anymore.

weyert:
@feywind Can you trigger a new Kokoro build so to verify my changes work with it?

feywind:
Yeah, it's set not to run unless the last commit was from one of the repo authors (for security).

I think we're still waiting on a resolution for the sinon thing, but it might be safe to just peg it at 10.0.0 for now. I'll look at that today. There are a few build weirdnesses I want to check on anyhow.

feywind:
The new error is this:
```
> nodejs-docs-samples-pubsub@ test /workspace/samples
> mocha system-test --timeout 600000


Error: Cannot find module '../package.json'
Require stack:
- /workspace/build/src/pubsub.js
- /workspace/build/src/index.js
- /workspace/samples/system-test/openTelemetryTracing.test.js
- /workspace/samples/node_modules/mocha/lib/esm-utils.js
- /workspace/samples/node_modules/mocha/lib/mocha.js
- /workspace/samples/node_modules/mocha/lib/cli/one-and-dones.js
- /workspace/samples/node_modules/mocha/lib/cli/options.js
- /workspace/samples/node_modules/mocha/lib/cli/cli.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:815:15)
    at Function.Module._load (internal/modules/cjs/loader.js:667:27)
```

But rather than going back and forth on it, let me take a look today and see if I can get it cleaned up.

weyert:
> The new error is this:
> 
> ```
> > nodejs-docs-samples-pubsub@ test /workspace/samples
> > mocha system-test --timeout 600000
> 
> 
> Error: Cannot find module '../package.json'

Cool! I think we maybe should avoid using `package.json`. I have pushed a draft PR for you to look in. I think it's easier to just generate a `version.ts` that contains the package version.

feywind:
We talked over the package.json thing, I think that's probably not going to change for now, but we might do something to make it more abstracted/standardized. I'd like to do it separately if we can get a clean build on this, though.

weyert:
> We talked over the package.json thing, I think that's probably not going to change for now, but we might do something to make it more abstracted/standardized. I'd like to do it separately if we can get a clean build on this, though.

That makes totally sense!  :)

package.json
```diff
@@ -38,7 +38,7 @@
     "clean": "gts clean",
     "compile": "tsc -p . && cp -r protos build/",
     "compile-protos": "compileProtos src",
-    "prepare": "npm run compile-protos && npm run compile && rm ./build/package.json",
+    "prepare": "npm run compile-protos && npm run compile",
```
weyert: I think you would need to revert the change from `require` to `import` approach for loading `package.json`. I don't think removing this is enough

feywind:
I'm a little bit confused, this branch now works and builds for me just fine locally. Actually removing the `package.json` in `build` broke some things for me. It's getting late here though, I'm going to look more next week.

weyert:
> I'm a little bit confused, this branch now works and builds for me just fine locally. Actually removing the `package.json` in `build` broke some things for me. It's getting late here though, I'm going to look more next week.

Most important that it works now :D Yeah confusing indeed.

weyert:
Any news?

feywind:
Not yet, sorry. I'm going to just retry it again for good measure, in case it was a (long lasting) transient error. I don't think anything is wrong with the code at this point.

feywind:
@weyert On my local machine, I did manage to get this part of the process to fail, but it was a different failure. I'm not sure why `package.json` would be required in `files` now, when it wasn't before. But I still don't get the Kokoro error, and I'm starting to wonder if someone tweaked settings somewhere. I'm asking others to see if they have any ideas.

weyert:
> @weyert On my local machine, I did manage to get this part of the process to fail, but it was a different failure. I'm not sure why `package.json` would be required in `files` now, when it wasn't before. But I still don't get the Kokoro error, and I'm starting to wonder if someone tweaked settings somewhere. I'm asking others to see if they have any ideas.

If you run it locally, what kind of output does the issues `npm pack`-command give you? The problem I had last week that something caused it not to bundle all files, see the problematic output below. If you look closely it's not including the whole `build`-directory only `build/package.json` is included. That's why it can't find the type definitions. Does this happen to you too?

```
npm notice 
npm notice 📦  @google-cloud/pubsub@2.10.0
npm notice === Tarball Contents === 
npm notice 11.4kB LICENSE           
npm notice 3.6kB  build/package.json
npm notice 3.2kB  package.json      
npm notice 57.5kB CHANGELOG.md      
npm notice 20.2kB README.md         
npm notice === Tarball Details === 
npm notice name:          @google-cloud/pubsub                    
npm notice version:       2.10.0                                  
npm notice filename:      google-cloud-pubsub-2.10.0.tgz          
npm notice package size:  22.0 kB                                 
npm notice unpacked size: 95.8 kB                                 
npm notice shasum:        3c77e1e46cfcbccb3cbd2de732d81a6fc5ff7823
npm notice integrity:     sha512-3H4qqXZWOSogu[...]6ERB1uWfewv2g==
npm notice total files:   5                                       
npm notice 
google-cloud-pubsub-2.10.0.tgz
```

weyert:
What we could try @feywind is to revert basically this change: https://github.com/googleapis/nodejs-pubsub/pull/1234/commits/0c2c76f8efa66e5f66d97430964fd30ac7e19261 and see what Kokoro things of it all.

feywind:
!! 🎉 
@weyert @aabmass Thank you both for all the work on this.

weyert:
Awesome! Thank you for merging :D

PR #1251, opened 2021-04-09
Draft: Add version command

weyert:
@feywind This is the draft.

I am not sure what your publishing process is but the idea is to run `npm run version` (should get run as part of npm publish) which will regenerate the `version.ts` file and populate it with the `version`-property of `package.json`.


feywind:
@weyert We talked about this a bit, and I think it's probably vetoed for now, sorry. :| It's mostly because so many of our packages do it the other way, so if we're going to change how it works, there's a desire to do it upstream in automation.

weyert:
All fine, happy the other PR problem with Kokoro got sorted :)

PR #1279, opened 2021-05-05
feat: update publisher options all the way through the topic object tree

### LINKED ISSUE #1103: topic.setPublishOptions fails to change batching settings
#### Environment details

  - OS: Linux
  - Node.js version: 12.x
  - npm version: 6.14.5
  - `@google-cloud/pubsub` version: 2.5.0

#### Steps to reproduce

  1. Create a topic
  2. Call `setPublishOptions` on the topic to change the batching settings
  3. Call `publish` or `publishJSON` on the topic to send a message
  4. Observe that new batching settings are not used

Probable root cause: `publisher.setOptions` doesn't propagate the new batching settings to its `this.queue` / `this.orderedQueues`: https://github.com/googleapis/nodejs-pubsub/blob/master/src/publisher/index.ts#L245

Workaround: always include batching options when creating the topic object.

feywind:
This also adds a defaults getter (`getPublishOptionDefaults()`) for publishing options.

Fixes #1103 🦕🦖🐢

### DIFF
--- a/src/publisher/index.ts
+++ b/src/publisher/index.ts
@@ -206,14 +206,19 @@ export class Publisher {
       queue.resumePublishing();
     }
   }
+
   /**
-   * Sets the Publisher options.
+   * Returns the set of default options used for {@link Publisher}. The
+   * returned value is a copy, and editing it will have no effect elsehwere.
+   *
+   * This is a non-static method to make it easier to access/stub.
    *
    * @private
    *
-   * @param {PublishOptions} options The publisher options.
+   * @returns {PublishOptions}
    */
-  setOptions(options = {} as PublishOptions): void {
+  getOptionDefaults(): PublishOptions {
+    // Return a unique copy to avoid shenanigans.
     const defaults = {
       batching: {
         maxBytes: defaultOptions.publish.maxOutstandingBytes,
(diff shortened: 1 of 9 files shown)


feywind:
I'll look at the Kokoro ones tomorrow.

src/publisher/index.ts
```diff
@@ -236,14 +254,28 @@ export class Publisher {
 
     this.settings = {
       batching: {
-        maxBytes: Math.min(batching.maxBytes, BATCH_LIMITS.maxBytes!),
-        maxMessages: Math.min(batching.maxMessages, BATCH_LIMITS.maxMessages!),
-        maxMilliseconds: batching.maxMilliseconds,
+        maxBytes: Math.min(batching!.maxBytes!, BATCH_LIMITS.maxBytes!),
+        maxMessages: Math.min(
+          batching!.maxMessages!,
```
JustinBeckwith, line 259: All the `!` makes me nervous we're committing TypeScript abuse - are these truly nullable?

JustinBeckwith (approved):

src/publisher/index.ts
feywind: They're not in this particular case, unless `promisifyAll()` is breaking things for you (argh). It's using `extend()` to pull in a full tree of defaults. The original had the defaults inside the same method, so it was anonymously typed with non-nullable values.

Can we refer to these expressions where every value is followed by `!` as "the big bang"? >_>

PR #1262, opened 2021-04-16
samples: add schema samples, typescript sample support

feywind:
fixes: https://github.com/googleapis/nodejs-pubsub/issues/1220

The remainder of this PR is about adding schema sample snippets, as well as some preliminary TypeScript snippet support. There are still some linting issues with the generated samples, and I wanted to talk about the method used here before committing to it, so it's still marked as a work in progress.

PR #1296, opened 2021-05-21
tests: add system tests for schema additions

feywind:
This is another piece broken off from the work in: https://github.com/googleapis/nodejs-pubsub/pull/1262

### DIFF
--- a/system-test/pubsub.ts
+++ b/system-test/pubsub.ts
@@ -625,15 +625,18 @@ describe('pubsub', () => {
     it('should ack the message', done => {
       const subscription = topic.subscription(SUB_NAMES[1]);
 
-      let errored = false;
+      let finished = false;
       subscription.on('error', () => {
-        errored = true;
-        subscription.close(done);
+        if (!finished) {
+          finished = true;
+          subscription.close(done);
+        }
       });
       subscription.on('message', ack);
 
       function ack(message: Message) {
-        if (!errored) {
+        if (!finished) {
+          finished = true;
           message.ack();
           subscription.close(done);
         }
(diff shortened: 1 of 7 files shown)


system-test/pubsub.ts
```diff
@@ -1023,6 +1031,153 @@ describe('pubsub', () => {
...
+    });
+
+    it('should list schemas', async () => {
+      const schemaId = await setupTestSchema();
+
+      const basicList = await aiToArray(
+        pubsub.listSchemas(SchemaViews.Basic),
+        schemaId
+      );
+      assert.strictEqual(basicList.length, 1);
```
tmatsuo, line 1092: What about concurrent builds? Will it still suceeds?

system-test/pubsub.ts
```diff
@@ -1023,6 +1031,153 @@ describe('pubsub', () => {
...
+      assert.strictEqual(basicList[0].definition, '');
+
+      const fullList = await aiToArray(
+        pubsub.listSchemas(SchemaViews.Full),
+        schemaId
+      );
+      assert.strictEqual(fullList.length, 1);
```
tmatsuo, line 1099: ditto

system-test/pubsub.ts
```diff
@@ -1023,6 +1031,153 @@ describe('pubsub', () => {
...
+        await pubsub.validateSchema({
+          type: SchemaTypes.Avro,
+          definition: fakeSchemaDef,
+        });
+      } catch (e) {
+        assert.ok(e);
+      }
+    });
+
+    // The server doesn't seem to be returning proper responses for this.
```
tmatsuo, line 1147: Maybe you can file an issue for tracking?

tmatsuo (approved):

system-test/pubsub.ts
```diff
@@ -1023,6 +1031,153 @@ describe('pubsub', () => {
...
+    // Commenting out for now, until it can be discussed.
```
bcoe, line 1148: by convention, I would put this as a:

```
//  TODO(feywind): uncomment this test
```

In the future, I think we might even automatically open issues when folks do this.

system-test/pubsub.ts
feywind: Good question, though I think it will be okay, because `aiToArray` filters by schema ID. Those are using a unique ID generated per run.

PR #1311, opened 2021-06-11
tests: attempt to make some tests more deterministic

feywind:
This is an attempt to fix a few non-deterministic things that popped up from the schema sample tests.

### DIFF
--- a/samples/system-test/schema.test.ts
+++ b/samples/system-test/schema.test.ts
@@ -23,7 +23,7 @@ import {
   Topic,
 } from '@google-cloud/pubsub';
 import {assert} from 'chai';
-import {describe, it, afterEach} from 'mocha';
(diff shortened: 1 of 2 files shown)


samples/system-test/schema.test.ts
```diff
@@ -202,12 +202,13 @@ describe('schema', () => {
     assert.include(output, schema.id);
     assert.include(output, 'deleted.');
 
-    try {
+    // Because of caching delays, this can't be deterministic without a big wait.
+    /*try {
```
bcoe, line 206: Is this commented out intentionally?

bcoe (approved):

samples/system-test/schema.test.ts
feywind: (discussed internally)

PR #1333, opened 2021-06-22
samples: add samples for sending and receiving large messages using files

feywind:
Add samples for sending and receiving large messages, from and to files, respectively.

This is at the request of the Cloud console team, so that users have a way to manually pull large messages for testing.

I decided to switch the sample tests to TypeScript while I was adding new ones anyway, and I discovered an error in the test for dead letter queues, so that is also fixed in here.


anguillanneuf:
@feywind Would you provide the internal bug where the request is made? 

My concern is that writing messages to files isn't a Pub/Sub feature per se. But maybe I'm missing some context.

anguillanneuf:
I see it b/191182578. I believe there's some misunderstanding here. The `gcloud` method for pulling is not referring to the async pull but is equivalent to https://github.com/googleapis/nodejs-pubsub/blob/5127df489b08d29ea3ea5a1c71489c47f27c2ce4/samples/synchronousPull.js#L62

So this PR doesn't really address the issue in the original PR. Makes sense?

PR #1359, opened 2021-07-27
feat: add publisher side flow control

### LINKED ISSUE #1081: RFC: limit simutanous publish requests
Following on from this question: https://stackoverflow.com/questions/63286981/gcp-pubsub-why-does-publishing-200k-messages-rapidly-result-in-2-5-million-mess

If I've understood the issue correctly, basically when we throw lots of messages at the client/sdk rapidly it correctly splits them into batches but does not limit the number of simultaneous gRPC publish requests that are made.

These requests are (as far as I can tell) going over a single http2 connection, which in nodejs by default is limited to 100 concurrent streams

> peerMaxConcurrentStreams <number> Sets the maximum number of concurrent streams for the remote peer as if a SETTINGS frame had been received. Will be overridden if the remote peer sets its own value for maxConcurrentStreams. Default: 100.
ref: https://nodejs.org/api/http2.html

As far as I can tell in the instance of pubsub we are either:
* Taking this default of 100
* Reducing the value to 1/2 (based on `rg -uu max_concurrent_streams`)
* Or accepting a value set by the server side unknown to me

I believe that by issuing so many simultaneous requests that they get gummed up, exceed the deadline (or some other condition) and are automatically retried as per the retry behaviour specified here: https://github.com/googleapis/nodejs-pubsub/blob/master/src/v1/publisher_client_config.json#L13-L21

I'm not very familiar with the code base, especially the `google-gax` / `grpc-js` parts so I'm hoping someone with deeper knowledge can confirm this theory.

I think the problem can be solved by doing something along the lines of this:
https://github.com/googleapis/nodejs-pubsub/compare/master...mnahkies:mn/limit-num-of-inflight-publish-requests?expand=1

I haven't yet tested that solution, but plan to test this and if it fits in with the project direction can refine it and write unit tests etc

feywind:
This PR adds publisher side flow control support, which lets you throttle the amount of messages being sent by bytes or message count.

Work in progress, to get CI.


feywind:
Fixes: https://github.com/googleapis/nodejs-pubsub/issues/1081

src/topic.ts
```diff
@@ -797,6 +797,38 @@ export class Topic {
...
+   *
+   * @example
+   * const messageIdPromises: Promise<string>[] = [];
+   * for (let i = 0; i < 1000; i++) {
+   *   const [idPromise] = await topic.publishWithFlowControl(messageBuffer);
+   *   messageIdPromises.push(idPromise);
+   * }
+   * const ids = await Promise.all(messageIdPromises);
+   */
+  publishWithFlowControl(
```
kamalaboulhosn, line 825: I'm not sure `publishWithFlowControl` is the right name given they can also publish with the "error" flow control and that is still publishing with flow control. The coalescing Promises is kind of frustrating. I think it makes the method much more confusing. Confusing enough that it kind of makes me want to go back to the idea of having something like:

```
const messageIdPromises: Promise<string>[] = [];
for (let i = 0; i < 1000; i++) {
  await topic.readyForPublish(messageBuffer);
  const idPromise = topic.publish(messageBuffer);
  messageIdPromises.push(idPromise);
}
const ids = await Promise.all(messageIdPromises);
```

Maybe for Node, it doesn't make sense to have a "Pause" mode, just have "Ignore" (default) and "Error." If one is in "Error" mode, then one can avoid the error by waiting on the Promise returned by `readyForPublish`, which can guarantee that a slot is available. Thoughts?

src/publisher/flow-control.ts
```diff
@@ -0,0 +1,197 @@
...
+ */
+
+import * as defer from 'p-defer';
+
+/**
+ * @typedef PublisherFlowControlAction Options for flow control actions
+ *
+ * @property {number} Ignore Ignore all flow control; don't take any action
+ *     based on outstanding requests.
+ * @property {number} Pause When flow control limits are exceeded, clients
```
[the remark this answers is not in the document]
feywind: Yeah, I was a little bit weird with "block" simply because you really don't block in Node (it has a different meaning). I can make it block if we want to be consistent, though.

src/topic.ts
feywind: Yeah, I went back and forth on this a number of times. I agree this method is really confusing. :| But also, it's safer and easier from my experimentation to take care of the waiting down inside the library. I think I'd rather keep that structure, but I am totally open to new ideas about the method name and/or how to do the nested wait.

kamalaboulhosn: I just have a really strong aversion to the introduction of Array. It feels like a hack. Given the choice between making the user do it by having a separate method or having to use the Array because of coalescing promises, I think the former is much cleaner.

feywind:
Adding `owlbot:ignore` because this adds samples.

samples/system-test/topics.test.ts
```diff
@@ -100,6 +100,22 @@ describe('topics', () => {
...
+      .get({autoCreate: true});
+    const output = execSync(
+      `${commandFor('publishWithFlowControl')} ${topicNameThree}`
+    );
+    const receivedMessage = await _pullOneMessage(subscription);
+    assert.strictEqual(receivedMessage.data.toString(), 'test!');
+    assert.ok(output.indexOf('Published 1000 with flow control settings') >= 0);
+
+    // Junk any remaining published messages.
+    await subscription.seek(new Date(Date.now()));
```
JustinBeckwith, line 116: I think this is the same as `new Date()` right?

samples/system-test/topics.test.ts
```diff
@@ -100,6 +100,22 @@ describe('topics', () => {
...
+    const [subscription] = await pubsub
+      .topic(topicNameThree)
+      .subscription(subscriptionNameOne)
+      .get({autoCreate: true});
+    const output = execSync(
+      `${commandFor('publishWithFlowControl')} ${topicNameThree}`
+    );
+    const receivedMessage = await _pullOneMessage(subscription);
+    assert.strictEqual(receivedMessage.data.toString(), 'test!');
+    assert.ok(output.indexOf('Published 1000 with flow control settings') >= 0);
```
JustinBeckwith, line 113: This is where chai's `assert.include` is quite nice.  It will show proper diffs when the test failed instead of a failed boolean assertion.

src/publisher/index.ts
```diff
@@ -76,7 +83,12 @@ export class Publisher {
   settings!: PublishOptions;
   queue: Queue;
   orderedQueues: Map<string, OrderedQueue>;
+  flowControl: FlowControl;
+
   constructor(topic: Topic, options?: PublishOptions) {
+    this.flowControl = new FlowControl(
+      options?.flowControlOptions ?? flowControlDefaults
```
JustinBeckwith, line 90: I think `||` would be safer than `??` here.  If someone passed `''` or `0` (I have no idea why they would), the only difference is `||` would correctly use the flow defaults.

src/publisher/pubsub-message.ts
```diff
@@ -0,0 +1,83 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {google} from '../../protos/protos';
+
+/**
+ * Strings are the only allowed values for keys and values in message attributes.
+ */
+export interface Attributes {
```
JustinBeckwith, line 22: This type kinda exists as `Record<string, string>` https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeystype

src/publisher/pubsub-message.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  let size = 0;
+
+  if (msg.data) {
+    size += msg.data.length;
+  }
+  if (msg.attributes) {
+    const attrs = msg.attributes;
+    for (const key of Object.getOwnPropertyNames(attrs)) {
+      const val = key + (attrs[key] ?? '');
```
JustinBeckwith, line 76: Same question above about `??` - do we explicitly want to treat empty strings as defined here?

test/publisher/flow-control.ts
```diff
@@ -0,0 +1,99 @@
...
+describe('Flow Controller', () => {
+  const optionsDefault: fc.FlowControlOptions = {
+    maxOutstandingMessages: 5,
+    maxOutstandingBytes: 100,
+  };
+
+  it('does basic bookkeeping correctly', async () => {
+    const flowPause = new fc.FlowControl(optionsDefault);
+    await flowPause.willSend(10, 1);
+    assert.strictEqual((flowPause as any).bytes, 10);
```
JustinBeckwith, line 34: huh - why the `as any` casts?

test/publisher/flow-publisher.ts
```diff
@@ -0,0 +1,133 @@
...
+    this.flowControl.setOptions(options.flowControlOptions!);
+  }
+}
+
+describe('Flow control publisher', () => {
+  let publisher: Publisher;
+  let sandbox: sinon.SinonSandbox;
+
+  beforeEach(() => {
+    sandbox = sinon.createSandbox();
```
JustinBeckwith, line 44: You don't have to create the sandbox each time - you can create it once in the `before` (or just above assign it.  The `sandbox.restore()` does the resetting for you.

samples/system-test/topics.test.ts
feywind: Probably! :) I'm used to moment now...

samples/system-test/topics.test.ts
feywind: I thought we were trying to work ourselves away from chai and just use the built in asserts? I'm happy to use the chai ones!

src/publisher/pubsub-message.ts
feywind: I'm sliiiightly nervous about changing this at this point simply because it was moved from another file, and is already exposed to users indirectly. But it sounds like it should be pretty equivalent.

src/publisher/pubsub-message.ts
feywind: To be fair, the values are checked against not being strings before the code gets to this point, but since it's exposed to users, I figured it should try to deal with potential gross input. I guess one possibility here is to do the string type check here as well.

test/publisher/flow-control.ts
feywind: The `.bytes` is private on the class. This is poking in to check the counters. There's probably no terribly compelling reason not to expose this as a readonly getter, so I might do that.

src/publisher/pubsub-message.ts
feywind: Actually I was wrong, I marked this one `@private`. So I'm just going to `||` to avoid breaking anyone who was somehow slipping the values through already.

JustinBeckwith (approved):

PR #1371, opened 2021-08-10
fix: export publish and subscribe options

### LINKED ISSUE #1227: Export types for publish/subscriber options
Hi

It would be great if the types for SubscriberOptions and PublishOptions would be exported. Most of the other option types are exported properly, these seem to have been missed. Is there any particular reason they shouldn't be exported? 

If its not intentional, this seems to be rather straightforward. I do not know if or when I'll be able to get a CLA request through so I didn't open a PR yet.

```diff
diff --git c/src/index.ts w/src/index.ts
index 8320ea3..15389ac 100644
--- c/src/index.ts
+++ w/src/index.ts
@@ -90,7 +90,8 @@ export {
   TestIamPermissionsCallback,
   IAM,
 } from './iam';
-export {Attributes, PublishCallback} from './publisher';
+export {Attributes, PublishCallback, PublishOptions} from './publisher';
+export {BatchPublishOptions} from './publisher/message-batch';
 export {PublishError} from './publisher/publish-error';
 export {
   PageOptions,
@@ -114,7 +115,7 @@ export {
   SeekResponse,
   Snapshot,
 } from './snapshot';
-export {Message} from './subscriber';
+export {Message, SubscriberOptions} from './subscriber';
 export {
   PushConfig,
   SubscriptionMetadata,
```

*update*: added BatchPublishOptions as it is needed because its part of PublishOptions

feywind:
Fixes https://github.com/googleapis/nodejs-pubsub/issues/1227

Some of the options objects that are passed down deep don't have all of the types exposed, so users of TypeScript can't set them gracefully.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -90,7 +90,8 @@ export {
   TestIamPermissionsCallback,
   IAM,
 } from './iam';
-export {Attributes, PublishCallback} from './publisher';
+export {Attributes, PublishCallback, PublishOptions} from './publisher';
+export {BatchPublishOptions} from './publisher/message-batch';
 export {PublishError} from './publisher/publish-error';
 export {
   PageOptions,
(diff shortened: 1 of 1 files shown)


alexander-fenster (approved):
New export is normally considered a `feat:` IIRC. Other than that, LGTM!

PR #1402, opened 2021-09-24
fix: deprecate maxExtension in favour of maxExtensionMinutes

### LINKED ISSUE #1208: Default option `maxExtensionMinutes` is treated as seconds
The default options is set as 60 minutes:

https://github.com/googleapis/nodejs-pubsub/blob/ccebf72508ba9f410c539eedaa879ec6fa423c0b/src/default-options.ts#L29-L31

The value `60` is applied here:

https://github.com/googleapis/nodejs-pubsub/blob/74fc12e8d1e26939309748536bfecb751be0f079/src/lease-manager.ts#L188

However `maxExtension` is documented as being a value in seconds, defaulting to `60`:

https://github.com/googleapis/nodejs-pubsub/blob/74fc12e8d1e26939309748536bfecb751be0f079/src/lease-manager.ts#L39-L40

Arguably the default should be `1` and when that default is applied it needs to be modified by 60.

feywind:
Background: maxExtension was incorrectly interpreted as seconds in the LeaseManager. This means that if users were relying on client side lease extension in subscribers, they would not actually extend leases for 60 minutes, but only 60 seconds. Also if the user passed in values, they'd be interpreted incorrectly. This PR deprecates the old field in favour of a new one that's explicitly named like the defaults, but also converts the old value if needed.

Fixes: https://github.com/googleapis/nodejs-pubsub/issues/1208

### DIFF
--- a/src/lease-manager.ts
+++ b/src/lease-manager.ts
@@ -182,10 +182,22 @@ export class LeaseManager extends EventEmitter {
    * Sets options for the LeaseManager.
    *
    * @param {FlowControlOptions} [options] The options.
+   *
+   * @throws {RangeError} If both maxExtension and maxExtensionMinutes are set.
+   *
    * @private
    */
   setOptions(options: FlowControlOptions): void {
-    // Convert the old deprecated maxExtension to avoid breaking clients.
+    // Convert the old deprecated maxExtension to avoid breaking clients,
+    // but allow only one.
+    if (
+      options.maxExtension !== undefined &&
+      options.maxExtensionMinutes !== undefined
+    ) {
+      throw new RangeError(
+        'Only one of "maxExtension" or "maxExtensionMinutes" may be set for subscriber lease management options'
+      );
+    }
     if (
       options.maxExtension !== undefined &&
       options.maxExtensionMinutes === undefined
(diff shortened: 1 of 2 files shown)


anguillanneuf (approved):
LGTM. Looks like the Node.js [sample](https://cloud.google.com/pubsub/docs/pull#flow_control) just uses `maxMessages`. Can you think of any other places we need to update?

src/lease-manager.ts
```diff
@@ -182,10 +185,19 @@ export class LeaseManager extends EventEmitter {
    * @private
    */
   setOptions(options: FlowControlOptions): void {
+    // Convert the old deprecated maxExtension to avoid breaking clients.
+    if (
+      options.maxExtension !== undefined &&
```
JustinBeckwith, line 190: What happens if both are defined?

JustinBeckwith (approved):
LGTM with nit.

feywind:
> LGTM. Looks like the Node.js [sample](https://cloud.google.com/pubsub/docs/pull#flow_control) just uses `maxMessages`. Can you think of any other places we need to update?

I think this is maxMessages rather than maxExtension, should be unchanged.

PR #1415, opened 2021-11-03
chore(cloud-rad): Add code fencing

fhinkel:
Code examples need code fencing around them to distingish from
rich text for TSDoc.

Internally b/179483748

Script used:
https://github.com/fhinkel/cloud-rad-script/blob/main/fixExampleComments.js


feywind (approved):
Interesting! Is it at all meaningful to try to specify a language on the fences?

fhinkel:
It's legal in TSDoc, not sure it'll make it all the way into Devsite. But it shouldn't be necessary.

PR #1416, opened 2021-11-03
fix: set only the Publish method to have a timeout of 5000ms

### LINKED ISSUE #1414: Topic creation fails with DEADLINE_EXCEEDED error
We create topics like this
```ts
const client = new PubSub()
const topic = client.topic('topic-name')
await this.topic.create()
```

It was working fine before but started failing with DEADLINE_EXCEEDED error after we updated to the latest version 2.18.1. 
We enabled gRPC debug mode and saw that all gRPC requests are sent with grpc-timeout header close to 5sec. After 5 seconds we get the error.
```
te: trailers
:path: /google.pubsub.v1.Publisher/CreateTopic
:method: POST
content-type: application/grpc
user-agent: grpc-node-js/1.3.6
:authority: pubsub.googleapis.com:443
accept-encoding: identity
grpc-accept-encoding: identity,deflate,gzip
authorization: 
x-goog-request-params: 
grpc-timeout: 4997m
x-goog-api-client: gax/2.25.0 gapic/2.18.1 gl-node/16.2.0 grpc/1.3.6 gccl/2.18.1
```

We checked with older version 2.8.0 and the grpc-timeout header is equal to 60 seconds. And topics were created fine. And it seems it also takes arount 5sec to create a topic. Looks like smth was changed between theses two version. 
```
te: trailers
:path: /google.pubsub.v1.Publisher/CreateTopic
:method: POST
content-type: application/grpc
user-agent: grpc-node-js/1.2.9
:authority: pubsub.googleapis.com:443
accept-encoding: identity,gzip
grpc-accept-encoding: identity,deflate,gzip
grpc-timeout: 60000m
authorization: 
x-goog-request-params: 
x-goog-api-client: gax/2.10.3 gapic/2.10.0 gl-node/14.6.0 grpc/1.2.9 gccl/2.10.0
```

We tried to pass timeout attribute in topic.create method but it did not make any difference. Is there any way to increase/override the grpc-timeout value?

feywind:
Fixes: https://github.com/googleapis/nodejs-pubsub/issues/1414

This was applied to all methods, and should only have applied to Publish.


feywind:
@kamalaboulhosn Do you mind giving this a look-over to make sure I tweaked the right number this time? Thanks! :D

JustinBeckwith (approved):

kamalaboulhosn (approved):

PR #1436, opened 2021-11-24
docs: fix documentation for publishing a message with an ordering key

### LINKED ISSUE #1435: Documentation is incorrect for publishing a message with orderingKey
https://github.com/googleapis/nodejs-pubsub/blob/main/samples/publishOrderedMessage.js#L69

`enableMessageOrdering` should be updated to `messageOrdering` according to the PublishOptions interface here https://github.com/googleapis/nodejs-pubsub/blob/98840fcc38a05be07de6e7dc97d8a79616940df9/src/publisher/index.ts#L38

a88zach:
- [X] Ensure the tests and linter pass
- [X] Code coverage does not decrease (if any source code was changed)

Fixes #1435 🦕

### DIFF
--- a/samples/publishOrderedMessage.js
+++ b/samples/publishOrderedMessage.js
@@ -66,7 +66,7 @@ async function main(
 
     // Publishes the message
     const messageId = await pubSubClient
-      .topic(topicName, {enableMessageOrdering: true})
+      .topic(topicName, {messageOrdering: true})
       .publishMessage(message);
 
     console.log(`Message ${messageId} published.`);
--- a/samples/resumePublish.js
+++ b/samples/resumePublish.js
@@ -53,7 +53,7 @@ async function main(
 
     // Publishes the message
     const publisher = pubSubClient.topic(topicName, {
-      enableMessageOrdering: true,
+      messageOrdering: true,
     });
     try {
       const message = {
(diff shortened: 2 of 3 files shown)


a88zach:
cla signed

feywind:
@a88zach Sorry for the delay. I think the situation might be more complicated and less flattering than it looks at first glance. The `messageOrdering` field, as far as I can tell, despite being defined in interfaces, isn't actually used anywhere. It looks like that was replaced with server-side ordering into subscriptions. Does that sound right, @kamalaboulhosn ? There's also this, which isn't great:

```
    /**
     * Identifies related messages for which publish order should be respected.
     * If a `Subscription` has `enableMessageOrdering` set to `true`, messages
     * published with the same `orderingKey` value will be delivered to
     * subscribers in the order in which they are received by the Pub/Sub
     * system.
     *
     * **EXPERIMENTAL:** This feature is part of a closed alpha release. This
     * API might be changed in backward-incompatible ways and is not recommended
     * for production use. It is not subject to any SLA or deprecation policy.
     *
     * @name Message#orderingKey
     * @type {string}
     */
    this.orderingKey = message!.orderingKey!;
```

That's in subscription.ts.

a88zach:
@feywind, I'm going to close this one since the discussion has gone stale

jketcham:
For what it's worth, I just ran into this when looking into publishing ordered messages with the node library. [Examples](https://github.com/googleapis/nodejs-pubsub/blob/ac5e7ef8f8cf887a54eac27e8f88241036d9a8bf/samples/publishOrderedMessage.js#L69) use `enableMessageOrdering`, but type defines it as `messageOrdering`. Some resolution here at least on the type side would be nice.

PR #1491, opened 2022-02-15
tests: fix another typing issue with streams

feywind:
Either more typings changes happened, or I missed this one somehow.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
   ],
   "scripts": {
     "presystem-test": "npm run compile",
-    "system-test": "GRPC_TRACE=all GRPC_VERBOSITY=DEBUG npx mocha build/system-test --timeout 600000",
+    "system-test": "mocha build/system-test --timeout 600000",
     "samples-test": "cd samples/ && npm link ../ && npm install && npm test && cd ../",
     "test": "c8 mocha build/test --recursive",
     "lint": "gts check",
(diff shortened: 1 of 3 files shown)


feywind:
Unrelated: I'd kind of love to get some of these system tests running off the emulator...

package.json
```diff
@@ -28,7 +28,7 @@
   ],
   "scripts": {
     "presystem-test": "npm run compile",
-    "system-test": "mocha build/system-test --timeout 600000",
+    "system-test": "GRPC_TRACE=all GRPC_VERBOSITY=DEBUG npx mocha build/system-test --timeout 600000",
```
bcoe, line 31: the `npx` will cause the latest version of mocha from npm to always be used, which I"m slightly concerned with from a supply chain point of view.

What I might perhaps do instead is use [cross-env](https://www.npmjs.com/package/cross-env), so that system-tests keep working on Windows, and drop the npx:

```
cross-env GRPC_TRACE=all GRPC_VERBOSITY=DEBUG mocha build/system-test --timeout=6000000
```

bcoe (changes requested):

package.json
feywind: Yeah, sorry, I forgot to mention in chat that these particular diffs are just debug stuff I'll be removing. I'm trying to see if the DEADLINE_EXCEEDED error can be reproduced here.

bcoe (approved):

PR #1492, opened 2022-02-16
samples: add samples for subscription with filtering

### LINKED ISSUE #1029: Add samples for RetryPolicy and filtering
This is in response to these two issues:

https://github.com/googleapis/nodejs-pubsub/issues/974
https://github.com/googleapis/nodejs-pubsub/issues/976

It looks like the support is in for us to do this, but some samples could go a long way toward making it accessible.

feywind:
Adds TS and JS samples for creating a subscription with a filtering string.

Fixes https://github.com/googleapis/nodejs-pubsub/issues/1029 🦕


samples/createSubscriptionWithFiltering.js
```diff
@@ -0,0 +1,78 @@
...
+//   description: Creates a new subscription with filtering.
+//   usage: node createSubscriptionWithFiltering.js <topic-name-or-id> <subscription-name-or-id> <filter-string>
+
+// [START pubsub_create_subscription_with_filter]
+/**
+ * TODO(developer): Uncomment these variables before running the sample.
+ */
+// const topicNameOrId = 'YOUR_TOPIC_NAME_OR_ID';
+// const subscriptionNameOrId = 'YOUR_SUBSCRIPTION_NAME_OR_ID';
+// const filterString = 'YOUR_FILTER_STRING';
```
hongalex, line 38: Should we give an example here like we do in the [Java samples](https://cloud.google.com/pubsub/docs/samples/pubsub-create-subscription-with-filter#pubsub_create_subscription_with_filter-java)?

samples/typescript/createSubscriptionWithFiltering.ts
```diff
@@ -0,0 +1,74 @@
...
+//   description: Creates a new subscription with filtering.
+//   usage: node createSubscriptionWithFiltering.js <topic-name-or-id> <subscription-name-or-id> <filter-string>
+
+// [START pubsub_create_subscription_with_filter]
+/**
+ * TODO(developer): Uncomment these variables before running the sample.
+ */
+// const topicNameOrId = 'YOUR_TOPIC_NAME_OR_ID';
+// const subscriptionNameOrId = 'YOUR_SUBSCRIPTION_NAME_OR_ID';
+// const filterString = 'YOUR_FILTER_STRING';
```
anguillanneuf, line 34: Same comment as above.

anguillanneuf (approved):
One nit but LGTM otherwise. 

Just FYI, we are currently not tracking Typescript in SnippetTracker because they aren't published on c.g.c. so the region tags aren't being used.

samples/createSubscriptionWithFiltering.js
feywind: Yeah, it's unclear to me since I don't want to give the impression, compared to the other default strings, that there's a "correct" filter to put there. I might add the sample filter as a comment after the rest of it...?

samples/typescript/createSubscriptionWithFiltering.ts
feywind: Thanks!

Yeah, I'm aware about the lack of TS tracking right now... it's something I would like to figure out, but for now I just want the samples to exist for users to find, and so we have them for later if we do that.

PR #1505, opened 2022-03-21
fix: update grpc.max_metadata_size to 4MiB for exactly-once, and shift ack/modack errors to 'debug' stream channel

### LINKED ISSUE #1504: Increase "grpc.max_metadata_size" GRPC parameter to 4 MB
This is needed to fix potential crashes when enabling exactly-once delivery on the server side.

b/225882734

feywind:
As the title says, this sets a default value for `grpc.max_metadata_size`.

While fixing that, I discovered that we are incorrectly raising a client error for ack/modack failures. Rather than making these go silent for users (for debug purposes), I moved them to a 'debug' channel that can also be subscribed on Subscription instances.

Fixes https://github.com/googleapis/nodejs-pubsub/issues/1504 🦕

### DIFF
--- a/src/message-queues.ts
+++ b/src/message-queues.ts
@@ -141,7 +141,10 @@ export abstract class MessageQueue {
     try {
       await this._sendBatch(batch);
     } catch (e) {
-      this._subscriber.emit('error', e);
+      // These queues are used for ack and modAck messages, which should
+      // never surface an error to the user level. However, we'll emit
+      // them onto this debug channel in case debug info is needed.
+      this._subscriber.emit('debug', e);
     }
 
     this.numInFlightRequests -= batchSize;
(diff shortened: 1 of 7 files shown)


pradn (commented):
Looks good. Please test:
1) subscribing for a long time from a exactly-once delivery-enabled subscription
2) subscribing from a regular subscription and turning on exactly-once for the subscription

In both cases, the subscribing should continue as normal.

feywind:
> Looks good. Please test:
> 
> 1. subscribing for a long time from a exactly-once delivery-enabled subscription
> 2. subscribing from a regular subscription and turning on exactly-once for the subscription
> 
> In both cases, the subscribing should continue as normal.

These are tested and pass after the fix for the ack/modack errors.

bcoe (approved):
Left one nit that might be worth fixing.

PR #1561, opened 2022-05-12
fix: detect subscription properties and warn for exactly-once

feywind:
This catches the subscription properties from the first delivered message, and warns the user if exactly-once is enabled. This is a feature in progress for the Node library, so the exactly-once features will not be available yet, and incorrectly used semantics may cause problems with an exactly-once subscription.

### DIFF
--- a/src/subscriber.ts
+++ b/src/subscriber.ts
@@ -30,6 +30,7 @@ import {Subscription} from './subscription';
 import {defaultOptions} from './default-options';
 import {SubscriberClient} from './v1';
 import {createSpan} from './opentelemetry-tracing';
+import {Throttler} from './util';
 
 export type PullResponse = google.pubsub.v1.IStreamingPullResponse;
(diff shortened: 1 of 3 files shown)


hongalex:
LGTM, but should get approval from @mmicatka or @pradn too

src/subscriber.ts
```diff
@@ -268,6 +273,26 @@ export class Subscriber extends EventEmitter {
     this._subscription = subscription;
     this.setOptions(options);
   }
+
+  /**
+   * Sets our subscription properties from the first incoming message.
```
pradn, line 278: `exactlyOnceDeliveryEnabled` can be changed any time, so we should check every message, not just the first one. You'd need some sort of throttle on the error, ofc.

pradn (changes requested):

feywind:
Pushing this through to get past the 16 test catch-22.

PR #1564, opened 2022-05-17
build!: update library to use Node 12

sofisl:
Migrate library to Node 12


feywind:
I talked it over with @sofisl , I think we're going to hold this back just a bit until I can figure out the timing with the Pub/Sub team.

test/subscription.ts
```diff
@@ -330,18 +330,19 @@ describe('Subscription', () => {
...
     it('should pass back any errors that occurs', done => {
       const fakeErr = new Error('err');
 
       sandbox.stub(subscriber, 'close').rejects(fakeErr);
 
-      subscription.close(err => {
-        assert.strictEqual(err, fakeErr);
-        done();
-      });
+      subscription?.close!(),
```
feywind, line 341: I think for whatever reason, the `?.` here is failing, so the test never concludes. I'm unsure why it would be undefined here and not before, but it's probably a more discoverable test failure if it has an undefined object exception anyhow (i.e. I'd just switch `subscription` back to `subscription!` for the Partial).

alexander-fenster (approved):

alexander-fenster (approved):

PR #1572, opened 2022-05-31
feat: add support for exactly once subscriptions

### LINKED ISSUE #1571: Add support for exactly-once subscriptions (EOS)
This is the overall issue to cover the effort to add EOS to the Node library.

feywind:
Adds support for exactly-once delivery subscriptions. Please see the samples for information on how to interact with exactly-once subscriptions properly (specifically, using the `*WithResponse()` methods).

Other client library folks - Mahesh needs to review this, so please don't merge until that happens.

Fixes https://github.com/googleapis/nodejs-pubsub/issues/1571 🦕


feywind:
Should be able to keep this open now that it has distinct commits.

src/message-stream.ts
```diff
@@ -151,6 +152,23 @@ export class MessageStream extends PassThrough {
     );
     this._keepAliveHandle.unref();
   }
+
+  /**
+   * Updates the stream ack deadline with the server.
+   *
+   * @param {Duration} deadline The new deadline value to set.
+   */
+  setAckDeadline(deadline: Duration) {
```
maheshgattani, line 161: setStreamAckDeadline?

src/subscriber.ts
```diff
@@ -30,13 +30,40 @@ import {Subscription} from './subscription';
...
+  FailedPrecondition: 'FAILED_PRECONDITION' as const,
+  Success: 'SUCCESS' as const,
+  Invalid: 'INVALID' as const,
+  Other: 'OTHER' as const,
+};
+export type AckResponse = ValueOf<typeof AckResponses>;
+
+/**
+ * Thrown when an error is detected in an ack/nack/modack call, when
+ * exactly-once is enabled on the subscription. This will only be thrown
```
maheshgattani, line 51: exactly-once delivery

src/subscriber.ts
```diff
@@ -184,6 +211,25 @@ export class Message {
     }
   }
 
+  /**
+   * Acknowledges the message, expecting a response (for exactly once subscriptions).
```
maheshgattani, line 215: exactly-once delivery

src/subscriber.ts
```diff
@@ -184,6 +211,25 @@ export class Message {
...
+   *
+   * @example
+   * ```
+   * subscription.on('message', async (message) => {
```
maheshgattani, line 219: Can we document what happens when a non exactly once sub calls this method? I assume that the promise should always return success.

src/subscriber.ts
```diff
@@ -184,6 +211,25 @@ export class Message {
...
+   *   const response = await message.ackWithResponse();
+   * });
+   * ```
+   */
+  async ackWithResponse(): Promise<AckResponse> {
+    if (!this._handled) {
```
maheshgattani, line 225: Can we short-circuit this in case exactly once delivery is disabled and return success?

src/subscriber.ts
```diff
@@ -196,6 +242,20 @@ export class Message {
...
   }
 
+  /**
+   * Modifies the ack deadline, expecting a response (for exactly once subscriptions).
+   *
+   * @param {number} deadline The number of seconds to extend the deadline.
+   * @private
+   */
+  async modAckWithResponse(deadline: number): Promise<AckResponse> {
+    if (!this._handled) {
```
maheshgattani, line 252: Same point as above: Can we short-circuit this in case exactly once delivery is disabled and return success?

src/subscriber.ts
```diff
@@ -212,6 +272,26 @@ export class Message {
...
+   *
+   * @example
+   * ```
+   * subscription.on('message', async (message) => {
+   *   const response = await message.nackWithResponse();
+   * });
+   * ```
+   */
+  async nackWithResponse(): Promise<AckResponse> {
+    if (!this._handled) {
```
maheshgattani, line 288: Same point as above: Can we short-circuit this in case exactly once delivery is disabled and return success?

src/subscriber.ts
```diff
@@ -212,6 +272,26 @@ export class Message {
...
     }
   }
+
+  /**
+   * Removes the message from our inventory and schedules it to be redelivered,
+   * with the modAck response being returned (for exactly once subscriptions).
+   *
+   * @example
+   * ```
+   * subscription.on('message', async (message) => {
```
maheshgattani, line 282: Same point as above: can we document the behavior for non exactly once subs?

src/subscriber.ts
```diff
@@ -196,6 +242,20 @@ export class Message {
     }
   }
 
+  /**
+   * Modifies the ack deadline, expecting a response (for exactly once subscriptions).
+   *
+   * @param {number} deadline The number of seconds to extend the deadline.
```
maheshgattani, line 248: Same point as above: can we document the behavior for non exactly once subs?

src/subscriber.ts
```diff
@@ -371,21 +449,16 @@ export class Subscriber extends EventEmitter {
...
-        console.error(
-          'WARNING: Exactly-once subscriptions are not yet supported ' +
-            'by the Node client library. This feature will be added ' +
-            'in a future release.'
-        )
-      );
-    }
-
     // Update ackDeadline in case the flag switched.
     if (previouslyEnabled !== this.isExactlyOnce) {
```
maheshgattani, line 453: Can we rename this variable to be isExactlyOnceDeliveryEnabled?

src/subscriber.ts
```diff
@@ -425,18 +498,42 @@ export class Subscriber extends EventEmitter {
...
+   *
+   * @param {Message} message The message to acknowledge.
+   * @returns {Promise<AckResponse>}
+   * @private
+   */
+  async ackWithResponse(message: Message): Promise<AckResponse> {
+    const ackTimeSeconds = (Date.now() - message.received) / 1000;
+    this.updateAckDeadline(ackTimeSeconds);
+
+    await this._acks.add(message);
```
maheshgattani, line 530: Sorry my experience with JS is limited: is this a blocking call? I assume no because otherwise returning a promise wouldn't make sense, but I just want to confirm.

samples/exactlyOnceListen.js
```diff
@@ -0,0 +1,101 @@
...
+    messageCount++;
+
+    // Use `ackWithResponse()` instead of `ack()` to get a Promise that tracks
+    // the result of the acknowledge call. When exactly-once delivery is enabled
+    // on the subscription, the message is guaranteed not to be delivered again
+    // if the ack Promise resolves.
+    try {
+      // When the Promise resolves, the value is always AckResponses.Success,
+      // signaling that the ack was accepted.
+      await message.ackWithResponse();
```
maheshgattani, line 66: Can we explain for the users what happens if `ackWithResponse` is called without exactly once delivery? I assume that it will say that the promise will always return success.

src/message-queues.ts
```diff
@@ -189,6 +290,96 @@ export abstract class MessageQueue {
...
+   */
+  handleAckSuccesses(batch: QueuedMessages) {
+    // Everyone gets a resolve!
+    batch.forEach(({responsePromise}) => {
+      responsePromise?.resolve();
+    });
+  }
+
+  /**
+   * If we get an RPC failure of any kind, this will sort out the mess
```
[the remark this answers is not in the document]
feywind: Yeah, I realized when looking over this again that the ack/modack calls may actually be fighting with the built-in gax retries. For exactly once delivery, we may not want gax retrying if there's a transient error that needs the exponential backoff. I'm looking into that as part of this pass.

src/message-stream.ts
feywind: I figured since it's in MessageStream, it's probably understood, but there's no harm in being explicit about it :)

src/subscriber.ts
feywind: I went ahead and just did several project-wide searches for this.

src/subscriber.ts
feywind: The way JS handles async/await is a little confusing looking. Basically the interpreter turns any async method into a couple of blocks with continuations using then() and catch(), and so the whole method ends up returning a Promise. In this particular case, the await will cause it to return a Promise that resumes the function's execution when resolve is called.

samples/exactlyOnceListen.js
feywind: Yeah, it will. I like all the doc comments, I'll go back and try to improve that stuff.

feywind:
I'm tracking review changes over here: https://github.com/googleapis/nodejs-pubsub/pull/1624

src/subscriber.ts
feywind: I had to think about this, but I think you're right, it's actually returning the incorrect response for non-EOS cases.

samples/listenForMessagesWithExactlyOnceDelivery.js
```diff
@@ -0,0 +1,103 @@
...
+    try {
+      // When the Promise resolves, the value is always AckResponses.Success,
+      // signaling that the ack was accepted. Note that you may call this
+      // method on a subscription without exactly-once delivery, but it will
+      // always return AckResponses.Success.
+      await message.ackWithResponse();
+      console.log(`Ack for message ${message.id} successful.`);
+    } catch (e) {
+      // In all other cases, the error passed on reject will explain why. This
+      // is only for permanent failures; transient errors are retried for a while.
```
maheshgattani, line 72: "automatically" to avoid questions on what is a "while"?

samples/typescript/listenForMessagesWithExactlyOnceDelivery.ts
```diff
@@ -0,0 +1,99 @@
...
+    try {
+      // When the Promise resolves, the value is always AckResponses.Success,
+      // signaling that the ack was accepted. Note that you may call this
+      // method on a subscription without exactly-once delivery, but it will
+      // always return AckResponses.Success.
+      await message.ackWithResponse();
+      console.log(`Ack for message ${message.id} successful.`);
+    } catch (e) {
+      // In all other cases, the error passed on reject will explain why. This
+      // is only for permanent failures; transient errors are retried for a while.
```
maheshgattani, line 68: same point as above

src/message-queues.ts
```diff
@@ -189,6 +330,124 @@ export abstract class MessageQueue {
...
+
+  /**
+   * If we get an RPC failure of any kind, this will take care of deciding
+   * what to do for each related ack/modAck. Successful ones will have their
+   * Promises resolved, permanent errors will have their Promises rejected,
+   * and transients will be returned for retry.
+   *
+   * Note that this is only used for subscriptions with exactly-once
+   * delivery enabled, so short-circuit to a success resolve on errors
+   * isn't handled here.
```
maheshgattani, line 354: I can't tell where this short-circuit is happening? I assume it's deeper in the stack. Should it be up top?

maheshgattani (approved):

PR #1580, opened 2022-06-09
chore: add config for releasing from legacy-v2 branch

feywind:
Adds release machinery to the legacy-v2 branch, which will support Node 10, and get releases on the 2.x version line.


chingor13 (approved):

.github/workflows/ci.yaml
```diff
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [12, 14, 16]
+        node: [10, 12, 14, 16]
```
chingor13, line 12: This change you probably want in the legacy branch

chingor13 (changes requested):

feywind:
@chingor13 Do you mind giving this another look? I moved the build change out to legacy-v2.

https://github.com/googleapis/nodejs-pubsub/blob/legacy-v2/.github/workflows/ci.yaml

chingor13 (approved):

feywind:
Owlbot changes: do not want, for this. It basically reverts the whole change =_=

I wonder if we need to do something to keep it from doing that on every PR.

PR #1582, opened 2022-06-09
feat: add minExtension setting and plug it into EOS and ackDeadline settings

feywind:
This covers steps 2 and 1-2 of the implementation doc:

* Add min extension parameter
* Watch for EOS settings in the stream
* If the user hasn't specified a min lease extension, use 60s for EOS; if non-EOS, continue as before
* Use the updated lease values when sending modAcks


src/default-options.ts
```diff
@@ -26,6 +26,12 @@ export const defaultOptions = {
     // in bytes, with the default lease manager.
     maxOutstandingBytes: 100 * 1024 * 1024,
 
+    // The minimum length of time a message's lease will be extended by.
+    minAckDeadline: undefined,
```
mmicatka, line 30: Should this have a default value? The Java library has the `minAckDeadline` default to 10 seconds

src/default-options.ts
```diff
@@ -26,6 +26,12 @@ export const defaultOptions = {
...
+
+    // The maximum length of time a message's lease will be extended by.
+    maxAckDeadline: undefined,
```
mmicatka, line 33: Similar to above, the Java library has the `maxAckDeadline` default to 600 seconds

src/default-options.ts
feywind: The Node library doesn't seem to have a default, and I was trying to maintain the existing behaviour for non-EOS. This is the relevant bit:

https://github.com/googleapis/nodejs-pubsub/blob/b255af63bdb150ccf42d90e2c4d17e6e94082ae9/src/lease-manager.ts#L278

feywind: Oh, also to be clear - the Node library defaults ackDeadline to 10s, but if the user overrides it, that value is always used. There's no lower or upper bound on ackDeadline as it adjusts for 99th percentile. I gather Node is a bit of an outlier in this...

src/subscriber.ts
```diff
@@ -297,6 +371,11 @@ export class Subscriber extends EventEmitter {
         )
       );
     }
+
+    // Update ackDeadline in case the flag switched.
+    if (previouslyEnabled !== this.isExactlyOnce) {
```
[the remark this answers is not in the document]
feywind: Yeah, `isExactlyOnce` should be a caching alias for that, but I will double check.

src/subscriber.ts
```diff
@@ -279,16 +295,74 @@ export class Subscriber extends EventEmitter {
...
+   * Should not be called from outside this class, except for unit tests.
+   *
+   * @param {number} [ackTimeSeconds] The number of seconds that the last
+   *   ack took after the message was received. If this is undefined, then
+   *   we won't update the histogram, but we will still recalculate the
+   *   ackDeadline based on the situation.
+   *
+   * @private
+   */
+  updateAckDeadline(ackTimeSeconds?: number) {
```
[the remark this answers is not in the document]
feywind: Yeah I think you're right, this could use a break up.

src/default-options.ts
mmicatka: For Java, we keep track if user-values are used for the deadlines, and if they are not, we change to use the EOS 10/600 defaults.

src/temporal.ts
```diff
@@ -0,0 +1,96 @@
...
+        throw new Error(`Invalid unit in call to totalOf(): ${totalOf}`);
+    }
+  }
+
+  /**
+   * Creates a Duration from a DurationLike, which is an object
+   * containing zero or more of the following: hours, seconds,
+   * minutes, millis.
+   */
+  static from(durationLike: DurationLike): Duration {
```
sofisl, line 88: NB: If any property is set twice, this would make millis commulative. But, in that case, that would seem like a mistake to me. For example, {minutes: 60, hours: 1}, would make duration 2 hours, but that seems like an improper use of the durationLike object (it seems like it should just be 120 min or 2 hours). We could throw an error if two properties are set, but feel free to ignore if there's a use case.

test/subscriber.ts
```diff
@@ -290,6 +291,68 @@ describe('Subscriber', () => {
...
+      assert.strictEqual(subscriber.ackDeadline, 60);
+
+      subscriber.setOptions({
+        minAckDeadline: Duration.from({seconds: 10}),
+      });
+      fakeDeadline = 1;
+      subscriber.ack(message);
+
+      assert.strictEqual(subscriber.ackDeadline, 10);
+    });
```
sofisl, line 325: What 'time' does Duration.from actually affect? Is it the moment subscriber.ack is called? Feel free to ignore if this doesn't make sense, but it would be nice to have a test where the deadline matches your expectations of the actual time of when it should be called, for example, exactly a minute later, etc.

src/temporal.ts
feywind: The canonical answer for this should be in the tc39 proposal. The official Duration class is going to separately store the various DurationLike fields. The effect should be the same, but I can double-check.

src/subscriber.ts
```diff
@@ -209,27 +214,40 @@ export class Message {
...
  * @property {BatchOptions} [batching] Request batching options.
  * @property {FlowControlOptions} [flowControl] Flow control options.
  * @property {boolean} [useLegacyFlowControl] Disables enforcing flow control
  *     settings at the Cloud PubSub server and uses the less accurate method
  *     of only enforcing flow control at the client side.
  * @property {MessageStreamOptions} [streamingOptions] Streaming options.
  */
+export interface SubscriberOptions {
+  /** @deprecated Use minAckDeadline and maxAckDeadline. */
+  ackDeadline?: number;
```
mmicatka, line 238: What is the difference between `ackDeadline` and `minAckDeadline`/`maxAckDeadline`? If the user wanted an explicit `ackDeadline` would that be the same as `minAckDeadline = ackDeadline` and `maxAckDeadline = ackDeadline`?

feywind: Yes. Setting ackDeadline will also set min/maxAckDeadline, to match the previous behaviour.

src/temporal.ts
feywind: Yeah, this is correctly implementing the very tiny subset of the proposal that I'm implementing here: https://tc39.es/proposal-temporal/docs/duration.html

from:
```
This static method creates a new Temporal.Duration from another value. If the value is another Temporal.Duration object, a new object representing the same duration is returned. If the value is any other object, a Temporal.Duration will be constructed from the values of any years, months, weeks, days, hours, minutes, seconds, milliseconds, microseconds, and nanoseconds properties that are present. Any missing ones will be assumed to be 0.
```

totalOf:
```
Calculates the number of units of time that can fit in a particular Temporal.Duration. If the duration IS NOT evenly divisible by the desired unit, then a fractional remainder will be present in the result.
```

There's not a requirement on internal representation for these two methods. It mostly matters for stuff like when you ask for the specific units separately (which I'm not doing).

acocuzzo (approved):
LGTM

PR #1588, opened 2022-06-14
feat: add "with response" versions of ack/nack/modack

feywind:
This adds ackWithResponse(), nackWithResponse(), and modAckWithResponse(), along with Promises that will simply complete to Success for this step (doc step 4).


feywind:
@meleu Unfortunately not. I'm working on finishing up the code work to get full exactly once support, but the approvals are taking a while.

PR #1594, opened 2022-07-06
samples: create BigQuery subscription

kamalaboulhosn:


feywind (approved):
Thanks for doing this!

feywind:
Looks like this broke us: https://github.com/open-telemetry/opentelemetry-js/pull/3065

I'm going to pin it at 1.3.x for now.

feywind:
I'm unsure why the OwlBot code check isn't passing after it runs. I'm going to try to just bypass it.

PR #1621, opened 2022-09-14
samples: add samples for and tests for exactly-once

feywind:
More work in progress, should be the last one.

### DIFF
--- a/samples/typescript/exactlyOnceCreate.ts
+++ b/samples/typescript/exactlyOnceCreate.ts
@@ -1,0 +1,73 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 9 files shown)


samples/typescript/exactlyOnceListen.ts
```diff
@@ -0,0 +1,91 @@
...
+  const messageHandler = (message: Message) => {
+    console.log(`Received message ${message.id}:`);
+    console.log(`\tData: ${message.data}`);
+    console.log(`\tAttributes: ${message.attributes}`);
+    messageCount++;
+
+    // "Ack" (acknowledge receipt of) the message
+    message
+      .ackWithResponse()
+      .then(() => {
```
hongalex, line 58: Can we should make it clear what is being returned here? (AckResponse)

In addition, something like [the python sample's comment](https://github.com/googleapis/python-pubsub/blob/0fa30c186ffeab44918a24a6ea5d92fad41b074d/samples/snippets/subscriber.py#L676-L679) might be good here too.

samples/typescript/exactlyOnceCreate.ts
```diff
@@ -0,0 +1,73 @@
...
+  await pubSubClient
+    .topic(topicNameOrId)
+    .createSubscription(subscriptionNameOrId, {
+      enableExactlyOnceDelivery: true,
+    });
+  console.log(
+    `Created subscription ${subscriptionNameOrId} with exactly-once delivery.`
+  );
+  console.log(
+    'To process messages, remember to check the return value of ack().'
```
hongalex, line 55: We want to check the return value of `ackWithResponse()` right?

PR #1636, opened 2022-09-21
fix: wait for 'drain' as well

### LINKED ISSUE #1620: Topic.flush() returns immediately, before all items are published
`Topic.flush()` is promisified, so it returns either a Promise or takes a callback, depending on how it's called. This calls `Publisher.flush()`, which likewise can deal with Promises. However, `MessageQueue.publish()`, which `Publisher.flush()` calls, does _not_ support Promises, only callbacks. But `Publisher.flush()` is expecting it to support Promises, so no callback is passed down, and `Publisher.flush()` gets back undefined, which passes through as a resolve, and no wait happens.

feywind:
It's possible to end up in a situation where more publishes are pending when publish() is called, but they aren't actually waited for. This adds 'drain' event support to the main publisher queue and waits for 'drain' on all queues before calling it done.

Maybe fixes #1620 🦕

### DIFF
--- a/test/publisher/index.ts
+++ b/test/publisher/index.ts
@@ -349,9 +349,9 @@ describe('Publisher', () => {
           .callsFake((messages, callbacks, callback) => {
             // Simulate the drain taking longer than the publishes. This can
             // happen if more messages are queued during the publish().
-            setTimeout(() => {
+            process.nextTick(() => {
               publisher.queue.emit('drain');
-            }, 50);
+            });
             if (typeof callback === 'function') callback(null);
           });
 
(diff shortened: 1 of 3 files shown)


test/publisher/index.ts
```diff
@@ -347,6 +347,11 @@ describe('Publisher', () => {
         sandbox
           .stub(FakeQueue.prototype, '_publish')
           .callsFake((messages, callbacks, callback) => {
+            // Simulate the drain taking longer than the publishes. This can
+            // happen if more messages are queued during the publish().
+            setTimeout(() => {
+              publisher.queue.emit('drain');
+            }, 50);
```
alexander-fenster, line 354: This will fail eventually, I can bet :)  Maybe use longer timeout?

alexander-fenster (approved):

test/publisher/index.ts
feywind: This just needs to be on the next tick, really, so I could just change this to nextTick().

PR #1638, opened 2022-09-27
chore: revert problematic stuff from the v3 branch

feywind:
This is in preparation for getting the legacy-v2 branch working for releases.


feywind:
That little manoeuvre is going to cost us a few CVEs =_= I'm starting to think maybe this legacy branch isn't going to work but let's see what the CI does.

feywind:
I don't think this is going to end up being practical due to dependency chains.

PR #1666, opened 2022-12-16
samples: add more TS samples

feywind:
Convert more samples to TypeScript, to work with the typeless sample bot.

### DIFF
--- a/owlbot.py
+++ b/owlbot.py
@@ -31,10 +31,16 @@ node.typeless_samples_hermetic()
 # We need to run this before the main owlbot processing, to make
 # sure quickstart.js gets gts fixed before the README is generated.
 # This needs to be worked out more properly, this is temporary.
-logger.debug("Running prelint...")
-shell.run(["npm", "run", "prelint"], check=False)
+logger.debug("Copy eslint config")
+shell.run(
+    ["cp", "-r", f"/synthtool/node_modules", "."],
+    check=True,
+)
 logger.debug("Running fix...")
-shell.run(["npm", "run", "fix"])
+shell.run(
+    [f"/synthtool/node_modules/.bin/gts", "fix"],
+    check=False,
+)
 
 node.owlbot_main(templates_excludes=[
     'src/index.ts',
--- a/owlbot.py
+++ b/owlbot.py
@@ -28,6 +28,14 @@ os.chdir(old_path)
 
 node.typeless_samples_hermetic()
 
+# We need to run this before the main owlbot processing, to make
+# sure quickstart.js gets gts fixed before the README is generated.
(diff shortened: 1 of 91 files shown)


owlbot.py
```diff
@@ -17,19 +17,31 @@
...
-"""
 import os
 from synthtool import shell
 from synthtool.log import logger
 old_path = os.getcwd()
 os.chdir("/synthtool")
-logger.debug("Update typeless sample bot [1.1.0]")
-shell.run(["npm", "i", "@google-cloud/typeless-sample-bot@1.1.0"])
+logger.debug("Update typeless sample bot [1.3.0]")
+shell.run(["npm", "i", "@google-cloud/typeless-sample-bot@1.3.0"])
```
bcoe, line 26: I thought we built `@google-cloud/typeless-sample-bot` into the post-processor, why is this install necessary?

owlbot.py
```diff
@@ -17,19 +17,31 @@
...
 os.chdir(old_path)
-"""
 
 node.typeless_samples_hermetic()
 
+# We need to run this before the main owlbot processing, to make
+# sure quickstart.js gets gts fixed before the README is generated.
+# This needs to be worked out more properly, this is temporary.
+logger.debug("Copy eslint config")
+shell.run(
```
bcoe, line 35: `gts` should already be built into the container, see:

https://github.com/googleapis/synthtool/blob/master/synthtool/languages/node.py#L216

Which should be called by `owlbot_main` already.

bcoe (changes requested):
Left a couple comments. I think you should be able to simplify OwlBot.py, given that we have typeless sample bot installed already no?

owlbot.py
feywind: Mentioned in Karl's issue on the main bot repo, but I was basically saving review/release time looping through synthtool. It's my intent to get synthtool updated once it's fully stabilized, and I can remove this in another PR.

owlbot.py
feywind: Yeah, the trouble is that there's a chicken and egg problem that needs solving here :\ We need to gts fix the quickstart.js before the README gets generated by OwlBot. There are probably better ways to fix this, I'm happy to talk it over :)

bcoe: I wonder if we could change the order of operations in `synthtool` to solve this?

bcoe (approved):

owlbot.py
feywind: Mentioned in chat, but recorded here for posterity: because of the README generation step, the quickstart requires a gts pre-pass, but I'll work out how to fix this and come back with a PR removing the manual install and hopefully fixing that.

PR #1700, opened 2023-03-17
samples: add sample for subscription with retry policy

### LINKED ISSUE #1029: Add samples for RetryPolicy and filtering

feywind:
Adds TypeScript and JavaScript samples for making a subscription with a retry policy. Note that these are not part of the sample browser, so at this time they can only be found in the repo.

Fixes: https://github.com/googleapis/nodejs-pubsub/issues/1029


grayside (approved):
How will this be used within region tags?

sofisl (approved):

PR #1702, opened 2023-03-28
Flush is broken when ordering enabled and multiple batches present

kimjoar:
- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/nodejs-pubsub/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea

Fixes #<issue_number_goes_here> 🦕

---

`flush` is currently broken when you have ordering + multiple batches when the flush occurs,  as `publishDrain` gets called with a callback that's dropped in the `publish` implementation of `OrderedQueue`. The callback is not passed on to `beginNextPublish`, so  the flush will never complete as it's waiting for the callback to be called. Looks like this came in with https://github.com/googleapis/nodejs-pubsub/pull/1691.

Haven't looked into tests etc. Let me know what you think makes sense, if this is interesting to merge.

### DIFF
--- a/src/publisher/message-queues.ts
+++ b/src/publisher/message-queues.ts
@@ -303,12 +303,12 @@ export class OrderedQueue extends MessageQueue {
   /**
    * Starts a timeout to publish any pending messages.
    */
-  beginNextPublish(): void {
+  beginNextPublish(callback?: PublishDone): void {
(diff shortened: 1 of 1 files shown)


kimjoar:
I think I've realized that flush shouldn't be used at all, and breaks ordering, if I'm not mistaken. Not clear from the docs.

PR #1709, opened 2023-04-13
fix: handle receipt modAck and lease extensions with exactly-once delivery correctly

### LINKED ISSUE #1697: Exactly-once delivery should wait for the receipt modAck to succeed before proceeding
The current behaviour is that the receipt modAck is "fire and forget", much like non-EOD receipt modAcks.

However, any delivery to the client logic should block on a successful receipt modAck. If the receipt modAck fails with a permanent error, we should not send the message to the client and not even log it. If it's a temporary failure, it should be retried.

feywind:
Two fixes in this one:

- receipt modAcks were not being handled properly in exactly-once delivery mode; they were basically being treated the same as non-EOD receipt modAcks
- the leasing manager was not doing modAckWithResponse when extending leases, which could cause memory and CPU leaks over time for exactly-once delivery subscriptions

I also discovered that we were missing a bunch of unit tests around this stuff, so I added some more.

Fixes https://github.com/googleapis/nodejs-pubsub/issues/1697

### DIFF
--- a/src/lease-manager.ts
+++ b/src/lease-manager.ts
@@ -257,7 +257,15 @@ export class LeaseManager extends EventEmitter {
       const lifespan = (Date.now() - message.received) / (60 * 1000);
 
       if (lifespan < this._options.maxExtensionMinutes!) {
-        message.modAck(deadline);
+        if (this._subscriber.isExactlyOnceDelivery) {
+          message.modAckWithResponse(deadline).catch(() => {
+            // In the case of a permanent failure (temporary failures are retried),
+            // we need to stop trying to lease-manage the message.
+            this.remove(message);
+          });
+        } else {
+          message.modAck(deadline);
+        }
       } else {
         this.remove(message);
       }
--- a/src/lease-manager.ts
+++ b/src/lease-manager.ts
@@ -15,7 +15,7 @@
  */
 
 import {EventEmitter} from 'events';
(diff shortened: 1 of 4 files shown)


acocuzzo (approved):

src/lease-manager.ts
```diff
@@ -257,7 +257,15 @@ export class LeaseManager extends EventEmitter {
       const lifespan = (Date.now() - message.received) / (60 * 1000);
 
       if (lifespan < this._options.maxExtensionMinutes!) {
-        message.modAck(deadline);
+        if (this._subscriber.isExactlyOnceDelivery) {
+          message.modAckWithResponse(deadline).catch(() => {
+            // In the case of a permanent failure (temporary failures are retried),
+            // we need to stop trying to lease-manage the message.
+            this.remove(message);
```
maheshgattani, line 264: We should be setting the response for the eventual ack as failed as well. This way, when the client acks the message, we do not have to issue a server call. If the modack fails with a permanent error, ack will also fail with a permanent error.

feywind: @maheshgattani Can you check the change I just made to validate that it meets the requirements? The ack/modAck/nack methods will only throw a failure if the error is permanent, so I think this covers it.

maheshgattani (approved):

PR #1713, opened 2023-04-15
feat: rework low level message stream retries, add debugging

### LINKED ISSUE #1712: Improve streaming pull stream retry behaviour
There are several potentially problematic behaviours with streaming pull stream retries that may have been making some of the known issues worse (or causing them).

- The minimum retry time is too high
- The default maximum retry time is too high (5 minutes)
- The back-off behaviour could result in very long retry times with long-term streams
- Streams within the stream pool are retried all at once, which might increase lag

In order to better get debugging information on these in the future, we also want to add some more stream logging messages for when things disconnect and reconnect. These will go out to the 'debug' channel of the subscription.

feywind:
Adjusts stream retries to standard exponential backoff behaviour, adds a few knobs for tweaking the settings of the backoff, and adds some 'debug' events on the subscription object for seeing when disconnects and reconnects happen.

Fixes #1712

### DIFF
--- a/src/message-stream.ts
+++ b/src/message-stream.ts
@@ -25,6 +25,7 @@ import {google} from '../protos/protos';
 import {defaultOptions} from './default-options';
 import {Duration} from './temporal';
 import {ExponentialRetry} from './exponential-retry';
+import {DebugMessage} from './debug';
 
 /*!
  * Frequency to ping streams.
@@ -359,7 +360,7 @@ export class MessageStream extends PassThrough {
     if (PullRetry.retry(status)) {
       this.emit(
         'debug',
-        new Error(
+        new DebugMessage(
           `Subscriber stream ${index} has ended with status ${status.code}; will be retried.`
         )
       );
@@ -372,7 +373,7 @@ export class MessageStream extends PassThrough {
     } else if (this._activeStreams() === 0) {
       this.emit(
         'debug',
-        new Error(
+        new DebugMessage(
           `Subscriber stream ${index} has ended with status ${status.code}; will not be retried.`
         )
       );
--- a/src/message-stream.ts
+++ b/src/message-stream.ts
@@ -265,6 +265,10 @@ export class MessageStream extends PassThrough {
    * @returns {Promise}
    */
   private async _fillStreamPool(): Promise<void> {
+    if (this.destroyed) {
+      return;
+    }
+
     let client!: ClientStub;
 
     try {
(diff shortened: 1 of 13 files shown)


src/message-stream.ts
```diff
@@ -268,65 +321,90 @@ export class MessageStream extends PassThrough {
...
+        'debug',
+        new Error(
+          `Subscriber stream ${index} has ended with status ${status.code}; will be retried.`
+        )
+      );
+      if (PullRetry.resetFailures(status)) {
+        this._retrier.reset(this._streams[index]);
+      }
+      this._retrier.retryLater(this._streams[index], () => {
+        this._fillStreamPool();
```
kamalaboulhosn, line 370: So we are still kind of "batching" the retries of streams given that we are using `_fillStreamPool`, is that right? So if two streams fail, we will retry both of them at whatever point the soonest one wants to be retried, if I'm reading this correctly. I rather we do the retries of individual streams independently.

kamalaboulhosn (changes requested):

src/message-stream.ts
```diff
@@ -268,65 +321,90 @@ export class MessageStream extends PassThrough {
...
-      const delay = this._retrier.createTimeout();
-      this._fillHandle = setTimeout(() => this._fillStreamPool(), delay);
-    } else if (!this._streams.size) {
+  private _onEnd(index: number, status: grpc.StatusObject): void {
+    this._removeStream(index);
+
+    if (PullRetry.retry(status)) {
+      this.emit(
+        'debug',
+        new Error(
```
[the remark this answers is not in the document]
feywind: I admit it's a bit of an abuse of what Error is supposed to do, but basically when you `new Error`, it's just making an object with some messages and a stack trace. There shouldn't be any other implications there. If that turns out to be wrong (I don't think it is though) we can always patch these later since they're not "load bearing" API surface. (Just advisory things the user might log, for the most part.)

src/message-stream.ts
feywind: I would say that the appearance of batching is superficial at this point. I mostly just didn't want to completely restructure the code. This has changed so that _fillStreamPool just looks to see if any of the streams need replacing and does that. I guess in practice it could end up retrying something sooner than it's supposed to be retried, but all of the streams should be going to the same place anyway. (So if one succeeds, all will succeed.)

src/message-stream.ts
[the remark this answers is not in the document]
feywind: I went ahead and changed this around to be more explicitly advisory/loggy.

src/message-stream.ts
feywind: I went ahead and split the retry up into its own method, so they're fully independent now.

kamalaboulhosn (approved):

PR #1723, opened 2023-05-09
chore: pull promise-based publish out of OTel change

feywind:
This shifts the publisher to use Promise-based logic instead of callbacks. This was pulled out of the OpenTelemetry change to make that one simpler.

### DIFF
--- a/src/publisher/message-queues.ts
+++ b/src/publisher/message-queues.ts
@@ -152,19 +152,23 @@ export class Queue extends MessageQueue {
    */
   add(message: PubsubMessage, callback: PublishCallback): void {
     if (!this.batch.canFit(message)) {
-      // Ignore errors.
+      // Make a background best-effort attempt to clear out the
+      // queue. If this fails, we'll basically just be overloaded
+      // for a bit.
       this.publish().catch(() => {});
     }
 
     this.batch.add(message, callback);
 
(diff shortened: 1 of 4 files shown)


bcoe (approved):

src/publisher/message-queues.ts
```diff
@@ -309,7 +307,10 @@ export class OrderedQueue extends MessageQueue {
     const timeWaiting = Date.now() - this.currentBatch.created;
     const delay = Math.max(0, maxMilliseconds - timeWaiting);
 
-    this.pending = setTimeout(() => this.publish(), delay);
+    this.pending = setTimeout(() => {
+      // Ignore errors.
+      this.publish().catch(() => {});
```
bcoe, line 312: What's the motivation for ignoring errors?

bcoe (approved):
I like the first class support for promises better than using promisify 👍

src/publisher/message-queues.ts
feywind: Discussed in chat, but copied here for posterity: I was mostly just emulating the behaviour of the old code, since it doesn't pass an error callback, but this does still change semantics a bit. (Code exceptions won't bubble up now.) I'm going to see if I can make this a bit more closely scoped first.

PR #1732, opened 2023-05-31
fix: allow updating stream count after opening

### LINKED ISSUE #1722: Calling Subscription.setOptions won't update the maxStreams count
Due to other recent changes (https://github.com/googleapis/nodejs-pubsub/issues/1712), the MessageStream class will only decide how many streams to use when it's constructed, so this setting will only work if passed to `pubsub.subscription()`. `MessageStream.fillStreams()` needs to pay attention to those config changes.

feywind:
Fixes: https://github.com/googleapis/nodejs-pubsub/issues/1722

I recognize that the unit tests for ExponentialRetry dig into it too much, but I don't really want to redo them in this context.

### DIFF
--- a/src/exponential-retry.ts
+++ b/src/exponential-retry.ts
@@ -72,13 +72,25 @@ export interface RetryCallback<T> {
  */
 export class ExponentialRetry<T> {
   private _items = new Heap<RetriedItem<T>>(comparator<T>);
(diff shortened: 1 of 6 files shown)


kamalaboulhosn:
Let's hold off on this. I'm not sure this is a good thing for us to do.

PR #1814, opened 2023-09-05
fix: set grpc keepalive time|outs by default

feywind:
This turns on grpc keepalives by default. We think this may fix quite a few (sometimes silent) timeout issues, because there's an issue deeper down in the stack that can result in the client library not realizing that its connection has been lost.

### DIFF
--- a/src/pubsub.ts
+++ b/src/pubsub.ts
@@ -272,16 +272,16 @@ export class PubSub {
   private schemaClient?: SchemaServiceClient;
 
   constructor(options?: ClientConfig) {
-    options = Object.assign({}, options || {});
-
-    // Needed for potentially large responses that may come from using exactly-once delivery.
-    // This will get passed down to grpc client objects.
-    const maxMetadataSize = 'grpc.max_metadata_size';
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    const optionsAny = options as any;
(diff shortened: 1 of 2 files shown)


maitrimangal (approved):

src/pubsub.ts
```diff
@@ -272,16 +272,16 @@ export class PubSub {
...
-      optionsAny[maxMetadataSize] = 4 * 1024 * 1024; // 4 MiB
-    }
+    // Needed for potentially large responses that may come from using exactly-once delivery,
+    // as well as trying to work around silent connection failures.
+    //
+    // These will get passed down to grpc client objects. User values will overwrite these.
+    const grpcDefaults = {
+      'grpc.max_metadata_size': 4 * 1024 * 1024, // 4 MiB
+      'grpc.keepalive_time_ms': 300000, // 5 minutes
+      'grpc.keepalive_timeout_ms': 20000, // 20 seconds
```
hongalex, line 282: Did we standardize on this somewhere?

hongalex (changes requested):

src/pubsub.ts
hongalex: Apparently this is the client default, so this seems fine with me.

hongalex (approved):

PR #1659, opened 2022-11-28
feat: add support for OTel context propagation and harmonized spans (v3.x)

feywind:
Work in progress, converting the existing OTel support to use context propagation and more finely grained spans.

Fixes https://github.com/googleapis/nodejs-pubsub/issues/1389

PR #1833, opened 2023-09-20
feat: add support for OTel context propagation and harmonized spans

### LINKED ISSUE #1389: OpenTelemetry integration misses the point: we need to propagate the span/context!
The code where the Subscriber reads the span from the attributes just creates the span and then throws it away:
https://github.com/googleapis/nodejs-pubsub/blob/99107930bbed8e29834acb2347f65efc6a39958f/src/subscriber.ts#L517-L527

A good OpenTelemetry customer would then set the span to the current span in the `on("message")` handler, so that any subsequent processing is attached to the remote span.

I'll also file a Google support case.

#### Environment details

  - `@google-cloud/pubsub` version: master / v2.17.0

#### Steps to reproduce

  1. Enable OpenTelemetry in both publisher and subscriber using `{ enableOpenTelemetryTracing: true }` options on topic and subscription.
  2. Perform some processing that creates spans, either manually or (for example) do a Datastore Query or Put.

It can be observed that actions are linked to the subscriber span, as can also be seen in this screenshot (spans take <1ms because nothing is done inside):
Screenshot 2021-08-30 at 16 47 23

Relates to #1078.

feywind:
This PR adds support for OTel context propagation using the standard W3C headers as Pub/Sub attributes. It also harmonizes the spans we're using with the planned spans in other language libraries.

Fixes https://github.com/googleapis/nodejs-pubsub/issues/1389

(This is the 4.x version of https://github.com/googleapis/nodejs-pubsub/pull/1659 - I'll still be backporting to there, but the changes will go here first.)


feywind:
This needs a bit more work due to some design changes in the spans.

feywind:
@omBratteng Glad to see a practical use case, thanks for linking :D

omBratteng:
@feywind looking forward to this getting merged 😎 

If I understand the new changes, there's no attributes added to the receiver span, as it is expected that they are set on the sender side?
https://github.com/feywind/nodejs-pubsub/blob/86491230c918dac24eb3f4bf5be8b5005267f8aa/src/telemetry-tracing.ts#L254-L277

weyert:
> FYI here are the changes for the latest spec updates, mostly done: [feywind#4](https://github.com/feywind/nodejs-pubsub/pull/4)

Awesome! Exciting to start using it :)

Bishal18:
Hi @feywind, by when can i expect a stable release for this? It would be an extremely important piece which would help to avoid unnecessary patching from our end to achieve the context propagation.

feywind:
> Hi @feywind, by when can i expect a stable release for this? It would be an extremely important piece which would help to avoid unnecessary patching from our end to achieve the context propagation.

I completely agree with you. The best I can say for full stable release (which has to include doc updates) is "aiming for 2024Q1". I'm hoping to get a preview release much sooner.

omBratteng:
Looking forward to preview release

naseemkullah:
@feywind @kamalaboulhosn  is there anything we can do to help this get merged?
If not, can  a smaller incremental change where the w3c context is propagated be merged sooner?
The `googclient_OpenTelemetrySpanContext` is not very useful when working with other services that use w3c context.
Thanks!

feywind:
@naseemkullah The delay has been annoying for sure. The good news is, this is being reviewed in pieces and there will hopefully be a beta release soon.

feywind:
There's one more of these in the review queue right now:
https://github.com/feywind/nodejs-pubsub/pull/5

feywind:
@hongalex The sample needs updating, but this should more or less be the finished PR now. Thanks for all the review help.

feywind:
This is now testable in beta form. I installed in a test project like so:

`npm i --save @google-cloud/pubsub@otel-beta`

The resulting package.json looks like this:

```json
{
  "dependencies": {
    "@google-cloud/opentelemetry-cloud-trace-exporter": "^2.1.0",
    "@google-cloud/pubsub": "^4.3.3-otel-beta.1",
    "@opentelemetry/sdk-trace-node": "^1.23.0",
    "@opentelemetry/semantic-conventions": "^1.23.0"
  }
}
```

And then you can take a look at [this sample](https://github.com/feywind/nodejs-pubsub/blob/otel-4/samples/openTelemetryTracing.js) for how to use it from JS, or [this sample](https://github.com/feywind/nodejs-pubsub/blob/otel-4/samples/typescript/openTelemetryTracing.ts) for how to use it from TS.

We've got more reviewing to do, and merges aren't going to happen for at least a week, but this may at least make it easier to do some testing if you like.

Further updates (for now) will come in as `4.3.3-otel-beta.2`, etc, and will update the tag `otel-beta` on npm.

feywind:
The `otel-beta` tag has been updated to `4.5.0-otel-beta.2`, which is current as of today. This is pretty close now!

hongalex (approved):

PR #1855, opened 2023-11-17
feat: Attempted implemented graceful shutdown

jjm340:
- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/nodejs-pubsub/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea

Fixes #<issue_number_goes_here> 🦕

### DIFF
--- a/src/lease-manager.ts
+++ b/src/lease-manager.ts
@@ -17,6 +17,7 @@
 import {EventEmitter} from 'events';
 import {AckError, Message, Subscriber} from './subscriber';
 import {defaultOptions} from './default-options';
+import defer = require('p-defer');
 
 export interface FlowControlOptions {
   allowExcessMessages?: boolean;
@@ -65,6 +66,7 @@ export class LeaseManager extends EventEmitter {
   private _pending: Message[];
   private _subscriber: Subscriber;
   private _timer?: NodeJS.Timeout;
+  private _onDrain?: defer.DeferredPromise<void>;
   constructor(sub: Subscriber, options = {}) {
     super();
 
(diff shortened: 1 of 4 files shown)


feywind:
@jjm340 Thanks for opening a PR. I think I can see some ways we can simplify this, but my main concern is that it might be considered a breaking change. I'd say "let me talk to the team on Thursday" except it's going to be US Thanksgiving :) so many of them will be out. Anyhow, I'll get back to you.

jjm340:
> @jjm340 Thanks for opening a PR. I think I can see some ways we can simplify this, but my main concern is that it might be considered a breaking change. I'd say "let me talk to the team on Thursday" except it's going to be US Thanksgiving :) so many of them will be out. Anyhow, I'll get back to you.

@feywind Thanks!! I had the same thought about this being a breaking change and a colleague and I discussed putting this behind an option flag, but wanted feedback from Google on that. I believe one side effect is that if this approach is used and a message is neither ACKed or NACKed, then the promise would never resolve and keep the consumer open indefinitely.

feywind:
I'll probably be picking this up again soon - the delay was because there was a fair amount of discussion about the way this should be implemented, and that is wrapping up now.

jjm340:
@feywind I was wondering if I should be fix my conflicts and attempt to put this behind a setting or is this being worked on by the team?

feywind:
There's a provided feature going in soon: https://github.com/googleapis/nodejs-pubsub/pull/2068

Is there anything in your PR that isn't covered by that? If so, I can look at it before it goes to release.

PR #1861, opened 2023-12-14
fix: add option to disable emulator auth handling (temp fix)

feywind:
Add manual control via PubSub options for whether the library uses "emulator mode", aka setting up for insecure SSL credentials for testing. Also fixes the heuristic to work with TPC.


src/pubsub.ts
```diff
@@ -70,6 +70,7 @@ export type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
 
 export interface ClientConfig extends gax.GrpcClientOptions {
   apiEndpoint?: string;
+  useFakeCredentials?: boolean;
```
kamalaboulhosn, line 73: I added a comment on the bug, but we should seriously consider alternatives to just adding another property like this. If we do end up adding it, let's not call it "useFakeCredentials."

feywind:
@kamalaboulhosn Mind taking another look at this and see if you're okay with what I changed? Unfortunately this is now blocking TPC updates. :\

hongalex (approved):

PR #1865, opened 2023-12-20
samples: add sample for unwrapped push subscriptions

feywind:
As the title says.


samples/createPushSubscriptionNoWrapper.js
```diff
@@ -0,0 +1,79 @@
...
+ * For more information, see the README.md under /pubsub and the documentation
+ * at https://cloud.google.com/pubsub/docs.
+ */
+
+// sample-metadata:
+//   title: Create Push Subscription With No Wrapper
+//   description: Creates a new push subscription, but disables metadata wrapping.
+//   usage: node createPushSubscriptionNoWrapper.js <topic-name-or-id> <subscription-name-or-id>
+
+// [START pubsub_create_push_no_wrapper_subscription]
```
hongalex, line 32:
```suggestion
// [START pubsub_create_unwrapped_push_subscription]
```

samples/createPushSubscriptionNoWrapper.js
```diff
@@ -0,0 +1,79 @@
...
+      },
+    },
+  };
+
+  await pubSubClient
+    .topic(topicNameOrId)
+    .createSubscription(subscriptionNameOrId, options);
+  console.log(`Subscription ${subscriptionNameOrId} created.`);
+}
+// [END pubsub_create_push_no_wrapper_subscription]
```
hongalex, line 65:
```suggestion
// [END pubsub_create_unwrapped_push_subscription]
```

samples/createPushSubscriptionNoWrapper.js
```diff
@@ -0,0 +1,79 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This is a generated sample, using the typeless sample bot. Please
+// look for the source TypeScript sample (.ts) for modifications.
+'use strict';
+
+/**
+ * This application demonstrates how to perform basic operations on
```
hongalex, line 20: This looks copy pasted from previous sample, is this ok? It seems a bit generic

samples/system-test/subscriptions.test.ts
```diff
@@ -186,6 +186,25 @@ describe('subscriptions', () => {
...
+        topic.name
+      } ${subName}`
+    );
+    assert.include(output, `Subscription ${subName} created.`);
+    const sub = await pubsub.subscription(subName);
+    const [subInfo] = await sub.get();
+    assert(subInfo.name === fullSubName(subName));
+    const [meta] = await subInfo.getMetadata();
+    assert(meta.pushConfig!.noWrapper!.writeMetadata === true);
+    /*const [subscriptions] = await pubsub.topic(topic.name).getSubscriptions();
```
hongalex, line 204: Uncomment or remove

samples/createPushSubscriptionNoWrapper.js
```diff
@@ -0,0 +1,79 @@
...
+
+async function createPushSubscriptionNoWrapper(
+  topicNameOrId,
+  subscriptionNameOrId
+) {
+  const options = {
+    pushConfig: {
+      // Set to an HTTPS endpoint of your choice. If necessary, register
+      // (authorize) the domain on which the server is hosted.
+      pushEndpoint: `https://${pubSubClient.projectId}.appspot.com/push`,
```
hongalex, line 53: Should this be a variable that the user configures, similar to `topicNameOrId`?

samples/createPushSubscriptionNoWrapper.js
```diff
@@ -0,0 +1,79 @@
...
+ * This application demonstrates how to perform basic operations on
+ * subscriptions with the Google Cloud Pub/Sub API.
+ *
+ * For more information, see the README.md under /pubsub and the documentation
+ * at https://cloud.google.com/pubsub/docs.
+ */
+
+// sample-metadata:
+//   title: Create Push Subscription With No Wrapper
+//   description: Creates a new push subscription, but disables metadata wrapping.
```
hongalex, line 29: writeMetadata seems to be set to `true` in the sample. Is this comment taken from another sample? If so, maybe we should clarify using the language from the proto https://github.com/googleapis/googleapis/blob/87e73054cd98b8d4a3f516066c6530213362bef4/google/pubsub/v1/pubsub.proto#L966 and https://github.com/googleapis/googleapis/blob/master/google/pubsub/v1/pubsub.proto#L914-L917

samples/createPushSubscriptionNoWrapper.js
feywind: Yeah, that is basically boilerplate for all of the samples in the library.

samples/system-test/subscriptions.test.ts
feywind: I don't know why this is even in here.

samples/createPushSubscriptionNoWrapper.js
feywind: This is copied from the createPushSubscription.js sample, so I'm just following that one's lead. I am okay with changing it to be explicit if we want, too. It is slightly ood.

samples/createPushSubscriptionNoWrapper.js
feywind: Yeah it's the wrapping it's disabling, not the writing, but I think this might be leftover language from the pre-"disable wrapping" change.

feywind: I like the proto language better. I've updated it.

samples/createPushSubscriptionNoWrapper.js
feywind: Giving a few more seconds thought to this topic, when I first took the library, all of the samples were inside single monolithic files. So this text was at the top. When I split them into separate files, I just left that text in place in each one. I'm not against changing it, but it's probably not part of this project. Just maintaining consistency for it.

samples/createPushSubscriptionNoWrapper.js
```diff
@@ -0,0 +1,82 @@
...
+  subscriptionNameOrId
+) {
+  const options = {
+    pushConfig: {
+      // Set to an HTTPS endpoint of your choice. If necessary, register
+      // (authorize) the domain on which the server is hosted.
+      pushEndpoint: `https://${pubSubClient.projectId}.appspot.com/push`,
+      // When true, writes the Pub/Sub message metadata to
+      // `x-goog-pubsub-<KEY>:<VAL>` headers of the HTTP request. Writes the
+      // Pub/Sub message attributes to `<KEY>:<VAL>` headers of the HTTP request.
```
hongalex, line 56: This comment seems duplicated

hongalex: Nevermind, I was confused about Pub/Sub message metadata vs attributes. Clarified here https://cloud.google.com/pubsub/docs/payload-unwrapping

hongalex (approved):

PR #1874, opened 2024-01-26
docs: correct max messages

### LINKED ISSUE #1744: Subscription maxMessages documented incorrectly
The [docs](https://cloud.google.com/nodejs/docs/reference/pubsub/latest/pubsub/subscription) say:

> By default Subscription objects allow you to process 100 messages at the same time.

However, digging into the source code, I can see that [elsewhere](https://github.com/googleapis/nodejs-pubsub/blob/bc7e915d5afb6b0b95250324d219067928bd07f6/src/lease-manager.ts#L44) `maxMessages` is documented as being 1000:

```
 * @property {number} [maxMessages=1000] The desired number of messages to allow
 *     in memory before pausing the message stream.
 ```

And checking in [`src/default-options.ts`](https://github.com/googleapis/nodejs-pubsub/blob/bc7e915d5afb6b0b95250324d219067928bd07f6/src/default-options.ts), I can see that 1000 is indeed the correct number:

```
export const defaultOptions = {
  subscription: {
    // The maximum number of messages that may be queued for receiving,
    // with the default lease manager.
    maxOutstandingMessages: 1000,
```

So I think the docs should be updated. Happy to open a PR.

richardnias:

Fixes #1744  🦕


feywind (dismissed):

feywind:
Thank you for the PR, I'll queue this up behind a few others I've got building.

feywind (approved):
Re-approving for automerge.

PR #1931, opened 2024-06-06
samples: add samples for cloud storage buckets

### LINKED ISSUE #1742: Add samples for cloud storage connector
For adding Node samples for the following tags:

* pubsub_cloudstorage_subscription
  * Messages sent to a topic from a storage bucket changes
  * ref: 284016504
* pubsub_create_cloud_storage_subscription
  * Messages published to a topic populate a storage bucket
  * ref: 283308258

feywind:
Fixes: https://github.com/googleapis/nodejs-pubsub/issues/1742


feywind:
It's only one sample for now, this is good to go again.

samples/README.md
```diff
@@ -253,6 +254,25 @@ __Usage:__
...
+Demonstrates how to create a subscription with Cloud Storage.
+
+View the [source code](https://github.com/googleapis/nodejs-pubsub/blob/main/samples/createSubscriptionWithCloudStorage.js).
+
+[![Open in Cloud Shell][shell_img]](https://console.cloud.google.com/cloudshell/open?git_repo=https://github.com/googleapis/nodejs-pubsub&page=editor&open_in_editor=samples/createSubscriptionWithCloudStorage.js,samples/README.md)
+
+__Usage:__
+
+
+`node createSubscriptionWithCloudStorage.js <topic-name> <subscription-name> <bucket> <filenamePrefix> <filenameSuffix> <maxDuration>`
```
lahuang4, line 268: Nit: why do we write "topic-name" and "subscription-name" (kebab-case) but "filenamePrefix" and "maxDuration" (camelCase)? Perhaps we should make these consistent?

samples/createSubscriptionWithCloudStorage.js
```diff
@@ -0,0 +1,99 @@
...
+// const filenameSuffix = 'YOUR_FILENAME_SUFFIX';
+// const maxDuration = 60;
+
+// Imports the Google Cloud client library
+const {PubSub} = require('@google-cloud/pubsub');
+
+// Creates a client; cache this for further use
+const pubSubClient = new PubSub();
+
+async function createSubscriptionWithCloudStorage(
```
lahuang4, line 49: Maybe name this createCloudStorageSubscription instead of createSubscriptionWithCloudStorage? Same suggestion applies to the typescript sample.

samples/createSubscriptionWithCloudStorage.js
```diff
@@ -0,0 +1,99 @@
...
+      bucket,
+      filenamePrefix,
+      filenameSuffix,
+      maxDuration: {
+        seconds: maxDuration,
+      },
+    },
+  };
+
+  const [sub] = await pubSubClient
```
lahuang4, line 68: Optional: I noticed the BQ subscription sample doesn't capture the created sub in a variable: https://cloud.google.com/pubsub/docs/samples/pubsub-create-bigquery-subscription?hl=en#pubsub_create_bigquery_subscription-nodejs_javascript Should we consider making these consistent too? Same suggestion applies to the typescript sample.

lahuang4 (commented):
Looks good overall!

samples/README.md
feywind: Good catch!

samples/createSubscriptionWithCloudStorage.js
feywind: I picked createSubscriptionWithCloudStorage because the library has five others with that naming scheme from earlier. Unless there's a strong objection, I think I'd probably keep the current name for that reason. Actually, the sample tags are all over the place already, so I don't think I'd want to try to fix it in this PR.

* pubsub_create_cloud_storage_subscription (this one)
* pubsub_dead_letter_create_subscription (createSubscriptionWithDeadLetterPolicy)
* pubsub_create_subscription_with_exactly_once_delivery (createSubscriptionWithExactlyOnceDelivery)
* pubsub_create_subscription_with_filter (createSubscriptionWithFiltering)
* pubsub_enable_subscription_ordering (createSubscriptionWithOrdering)
* createSubscriptionWithRetryPolicy - doesn't seem to have sample tags at all (??)

Maybe for the purposes of this PR, I'll just update the function name to match the tag and leave the filename as-is.

samples/createSubscriptionWithCloudStorage.js
feywind: I don't feel strongly about it, but it looks like the other createSubscription samples in here don't capture it either, so I'll change it.

danieljbruce (approved):

PR #1880, opened 2024-02-04
propagate set options to LeaseManager

dermasmid:
Fixes #1879

PR #1954, opened 2024-07-10
fix: propagate set options to LeaseManager (from https://github.com/googleapis/nodejs-pubsub/pull/1880)

### LINKED ISSUE #1879: calling setOptions on subscriber does not take effect
if changing flowControl settings after calling .on they dont not get into effect

```
const subscription = pubSubClient.subscription(subscriptionNameOrId);
const messageHandler = message => {

  message.ack();
};

subscription.on('message', messageHandler);

// does not have any effect on flow control
subscription.setOptions({
flowControl: {
          maxMessages: 1,
          allowExcessMessages: true,
 }
})

```

feywind:
Bringing this over from https://github.com/googleapis/nodejs-pubsub/pull/1880 (dermasmid@) so I can work CI on it.

Fixes https://github.com/googleapis/nodejs-pubsub/issues/1879


hongalex (approved):

src/subscriber.ts
```diff
@@ -771,11 +771,22 @@ export class Subscriber extends EventEmitter {
...
       const {maxStreams = defaultOptions.subscription.maxStreams} =
         options.streamingOptions;
+
       options.streamingOptions.maxStreams = Math.min(
         maxStreams,
         this.maxMessages
       );
     }
+
+    if (this._acks) {
```
hongalex, line 781: Let's not expose these settings more than they already are, since we don't want users to be able to set these batching settings. Flow control is ok though

PR #1990, opened 2024-10-23
feat: add async helper for ordered async processing of messages

feywind:
Adds an AsyncHelper class to the Node Pub/Sub library that helps to process messages in order even with an async handler. For ordering queues especially, it was problematic to try to handle the case of ordered messages coming in and being processed by async code; async uses Promises, and these are basically always guaranteed not to be processed inline. This helper queues messages from the stream.on handler and lets each process before starting the next.


feywind:
This will be added into the library elsewhere.

carlswann:
@feywind Do you know where/when it will be added? Was literally just thinking about creating a PR for this

PR #1996, opened 2024-11-06
maxOutstandingMessages per stream

dermasmid:
the `maxOutstandingMessages` is set per stream so we should divide by num of streams, right?


feywind:
Yes. Unfortunately this is being held up by some discussions about what to do with legacy flow control compatibility.

hongalex:
To add slightly more color, dividing the flow control among the total amount of stream does not mean that the messages will be delivered evenly to each streaming pull stream. The total throughput could be limited in that case.

feywind:
But importantly for this PR, that change could cause unexpected side effects, which is why we haven't done it yet.

PR #2020, opened 2025-03-26
fix: type widening to match gapic, and two typing fixes

### LINKED ISSUE #1707: Change type Buffer into Uint8Array in publishMessage

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

 **Is your feature request related to a problem? Please describe.**
In `publishMessage` method of Topic, it checks if data is instance of Buffer.
Some of modern serialization library uses Uint8Array as return value instead of Buffer. Although we can change it into Buffer using `Buffer.from`, it will copy Arraybuffer, which causes performance issue.
Since Buffer is subclass of Uint8Array, I think it can be changed into `data && !(data instanceof Uint8Array)`.

 **Describe the solution you'd like**
Change src/publisher/index.ts L195 into `    if (data && !(data instanceof Uint8Array)) {`.

 **Describe alternatives you've considered**
We can change Uint8Array into Buffer using `Buffer.from`method, but it will 'copy' array and create new one.

 **Additional context**
.

feywind:
Fixes https://github.com/googleapis/nodejs-pubsub/issues/1707


sofisl (approved):

leahecole (approved):

leahecole:
To fix CI you might have to pin to an earlier version of promisify until you upgrade to node 18 - there was an accidental minor version bump that should have been a major so 4.0.0 is safe

feywind:
> To fix CI you might have to pin to an earlier version of promisify until you upgrade to node 18 - there was an accidental minor version bump that should have been a major so 4.0.0 is safe

Yeah, I've been using 18 for local testing, which is probably premature. Hopefully we'll get the 18 support sorted soon on this repo, anyhow.

PR #1963, opened 2024-09-06
fix: set MaxBytes for AckQueue

feywind:
This PR supersedes this one: https://github.com/googleapis/nodejs-pubsub/pull/1897
Thanks to @dermasmid for getting that started; I'm starting a new one to have write access for changes and CI.

This implements a fix to keep ack/modack batches under a size that the server can handle.

Fixes https://github.com/googleapis/nodejs-pubsub/issues/1864

PR #2027, opened 2025-04-16
fix: check batch size before re-queueing retried messages

### LINKED ISSUE #2026: Message retrying queues requests without respecting batch size limits

[]()

_

When there's a PubSub backend error and acks are automatically re-queued to retry, if the ack queue is large enough, it occasionally triggers an `INVALID_ARGUMENT: Request payload size exceeds the limit: 524288 bytes.` error.

It seems like the automatic message retry path doesn't trigger flushing before re-queueing messages when the max batch size would be exceeded, which can lead to more acks ending up batched in the same request than expected, potentially leading to the error above.

I'd expect the number of ack IDs in a single acknowledgment request to never exceed whatever is specified in the `maxMessages` batching option.

Documentation describes `maxMessages` in `BatchOptions` as controlling the maximum number of messages batched together in requests: https://googleapis.dev/nodejs/pubsub/latest/global.html#BatchOptions

Calling `Message.ack()`, `Message.ackWithResponse()`, `Message.nack()`, and `Message.nackWithResponse()`, seems to properly batch messages together based on the specified `maxMessages` as documented.

1kilobit:
Fixes #2026 🦕

This adds the same batch size-based flushing logic added to `MessageQueue.add()` in PR #1963 to `MessageQueue.handleRetry()`, which also queues into the next batch.


feywind:
@1kilobit Thanks for the PR, I just noticed that I missed this!

feywind (dismissed):

abbrowne126 (approved):

feywind (approved):

PR #2037, opened 2025-05-06
feat: add timeout option and graceful shutdown to Subscription.close()

feywind:
Implements a new `timeout` and `behavior` option (using `Duration`) for the `Subscription.close()` method. These options are on the SubscriberOptions/SubscriptionOptions passed in when opening a subscriber.

This provides more control over the shutdown process:
- The subscriber stream is closed immediately.
- If the behaviour is set to NackImmediately, then all buffered messages are nacked, and the timeout specifies how long to wait for them to go through.
- If the behaviour is set to WaitForProcessing, then the remaining message machinery continues until the timeout is almost up, and then it switches to NackImmediately.
- If the timeout is zero, everything closes down immediately, as quickly as possible.


feywind:
Migrated to https://github.com/googleapis/nodejs-pubsub/pull/2068 to get CI.

PR #2062, opened 2025-07-03
feat: ad-hoc, targeted debug logging

feywind:
This feature will add the ability to selectively enable "printf debugging" output for diagnostic introspection. This uses the new standard Google ad-hoc debug logging infrastructure introduced a few months back.

Samples:

> GOOGLE_SDK_NODE_LOGGING=pubsub:*

All hand-written logic on top of the auto-generated RPCs.

> GOOGLE_SDK_NODE_LOGGING=pubsub*

All logging related to Pub/Sub (including the RPCs).

> GOOGLE_SDK_NODE_LOGGING=*

All logging, including things like auth.

A fuller guide will be forthcoming.

### DIFF
--- a/src/message-stream.ts
+++ b/src/message-stream.ts
@@ -15,7 +15,7 @@
  */
 
 import {promisify} from '@google-cloud/promisify';
-import {ClientStub, grpc} from 'google-gax';
+import {ClientStub, GoogleError, grpc} from 'google-gax';
 import * as isStreamEnded from 'is-stream-ended';
 import {PassThrough} from 'stream';
 
@@ -452,7 +452,12 @@ export class MessageStream extends PassThrough {
     const tracker = this._streams[index];
     const receivedStatus =
       !tracker.stream || (tracker.stream && !tracker.receivedStatus);
-    logs.subscriberStreams.error('error on stream %i: %o', index, err);
+
+    // For the user-cancelled errors, we don't need to show those, we're handling
+    // notifying of us closing the stream elsewhere.
+    if ((err as GoogleError).code !== 1) {
+      logs.subscriberStreams.error('error on stream %i: %o', index, err);
+    }
 
     if (typeof code !== 'number' || !receivedStatus) {
       this.emit('error', err);
--- a/src/message-stream.ts
+++ b/src/message-stream.ts
@@ -15,7 +15,7 @@
  */
 
 import {promisify} from '@google-cloud/promisify';
-import {ClientStub, grpc, loggingUtils} from 'google-gax';
+import {ClientStub, grpc} from 'google-gax';
 import * as isStreamEnded from 'is-stream-ended';
 import {PassThrough} from 'stream';
 
@@ -26,6 +26,7 @@ import {defaultOptions} from './default-options';
 import {Duration} from './temporal';
 import {ExponentialRetry} from './exponential-retry';
 import {DebugMessage} from './debug';
+import {logs as baseLogs} from './logs';
 
 /**
  * Loggers. Exported for unit tests.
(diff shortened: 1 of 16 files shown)


src/subscriber.ts
```diff
@@ -763,6 +779,21 @@ export class Subscriber extends EventEmitter {
...
 
+    logs.ackNack.info(
+      'message (ID %s, ackID %s) ack',
+      message.id,
+      message.ackId,
+    );
+
+    if (ackTimeSeconds > this._99th) {
+      logs.slowAck.info(
+        'message (ID %s, ackID %s) ack took longer than the 99th percentile (%s s)',
```
hongalex, line 790: Since these are logs, we should probably be more explicit here. 99th percentile of what?

src/message-stream.ts
```diff
@@ -452,7 +452,12 @@ export class MessageStream extends PassThrough {
     const tracker = this._streams[index];
     const receivedStatus =
       !tracker.stream || (tracker.stream && !tracker.receivedStatus);
-    logs.subscriberStreams.error('error on stream %i: %o', index, err);
+
+    // For the user-cancelled errors, we don't need to show those, we're handling
+    // notifying of us closing the stream elsewhere.
+    if ((err as GoogleError).code !== 1) {
```
hongalex, line 458:
```suggestion
    if ((err as GoogleError).code !== grpc.status.CANCELLED) {
```

src/publisher/message-queues.ts
```diff
@@ -211,25 +243,26 @@ export class Queue extends MessageQueue {
...
   }
 
   /**
    * Cancels any pending publishes and calls _publish immediately.
    *
    * Does _not_ attempt to further drain after one batch is sent.
    */
-  async publish(): Promise<void> {
-    await this._publishInternal(false);
+  async publish(reason: string): Promise<void> {
```
sofisl, line 254: since you're adding a param, wouldn't you need to add a default value to make it backwards-compatible or make it optional?

src/publisher/message-queues.ts
```diff
@@ -211,25 +243,26 @@ export class Queue extends MessageQueue {
...
+    await this._publishInternal(false, reason);
   }
 
   /**
    * Cancels any pending publishes and calls _publish immediately.
    *
    * @emits Queue#drain when all messages are sent.
    */
-  async _publishInternal(fullyDrain: boolean): Promise<void> {
+  async _publishInternal(fullyDrain: boolean, reason: string): Promise<void> {
```
sofisl, line 263: same here, optional?

sofisl (commented):
same general comment about [private elements](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements)

src/subscriber.ts
```diff
@@ -787,7 +787,7 @@ export class Subscriber extends EventEmitter {
 
     if (ackTimeSeconds > this._99th) {
       logs.slowAck.info(
-        'message (ID %s, ackID %s) ack took longer than the 99th percentile (%s s)',
+        'message (ID %s, ackID %s) ack took longer than the 99th percentile of ack/nack time (%s s)',
```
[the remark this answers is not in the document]
feywind: Yeah, I couldn't think of a good wording here, I'll just go with yours.

src/publisher/message-queues.ts
feywind: Also, re: private methods, this library is pretty old and wasn't built to use them, so it would be pretty painful to add it now, imo. I'll definitely think about that going forward, but e.g. a lot of the unit tests assume that innards are accessible for mocking. I'd like to find a better way to do the unit tests too, but there are over 900 of them now 😅

hongalex (dismissed):

src/publisher/message-queues.ts
feywind: I went ahead and made those optional parameter changes.

sofisl (approved):

PR #2091, opened 2025-09-19
chore: remove docs check from CI, never works due to 429 errors from GitHub

### LINKED ISSUE #2090: Docs pre-check never works because of 429 errors
GitHub changed their quota limits for loading files by non-logged-in bots, and now linkinator basically never works. Most of our other repos have stopped running it as a PR pre-check, so I want to do the same here and make the red/green work, as well as automerge.

feywind:
Fixes https://github.com/googleapis/nodejs-pubsub/issues/2090

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -60,3 +60,15 @@ jobs:
           node-version: 18
       - run: npm install
       - run: npm run lint
+  docs:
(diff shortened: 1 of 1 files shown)


feywind:
Link to newer one: https://github.com/googleapis/nodejs-pubsub/pull/2114

PR #2112, opened 2025-11-10
fix: move off the remaining rc versions

feywind:
Some of the library dependencies were still pointing at rc versions. This mitigates that.


quirogas (approved):
LGTM! The failing `docs` test appears to be a red herring, as the links seem to be working.

PR #2118, opened 2025-12-09
fix(WaitForProcessing): Shutdown option `WaitForProcessing` never entered on subscription close

### LINKED ISSUE #2117: Shutdown option `WaitForProcessing` never entered on subscription close

Not available, this bug is part of the client shutdown behavior and not part of the pubsub API.

https://github.com/googleapis/nodejs-pubsub/blob/main/samples/closeSubscriptionWithTimeout.js#L60-L69

1. Create a new subscription with 
```js
closeOptions: {
      behavior: SubscriptionCloseBehaviors.WaitForProcessing,
      timeout: Duration.from({seconds: 10}),
    },
```
2. Let the message processing callback of that subscription perform work that takes a couple of seconds to finalize. Does not need to exceed the `closeOption` timeout.
3. Publish messages to the topic that this subscription is bound to.
4. Call `await subscription.close()` during message processing
5. You will see that the promise is returned immediately and the `WaitForProcessing` closeOption was never entered

There seems to be a bug in the implementation of the condition during subscription close that decides to enter the `waitForProcessing` behavior in https://github.com/googleapis/nodejs-pubsub/blob/abeeb6c600e540b282de8523d75c83d3147e273c/src/subscriber.ts#L1001-L1004. The access to `this._inventory.isEmpty` seems to be undefined and would need to be changed to `this._inventory.isEmpty()` as also suggested in the comment of the PR originally introducing this functionality: https://github.com/googleapis/nodejs-pubsub/pull/2068#pullrequestreview-3545265767

It seems that with this bug, the `waitForProcessing` graceful shutdown feature introduced with https://github.com/googleapis/nodejs-pubsub/pull/2068 is not working as expected and the shutdown behavior is never initiated.

The `WaitForProcessing` closeOption was recently introduced in version `5.2.0` with https://github.com/googleapis/nodejs-pubsub/pull/2068. In the supplied tests, the `isEmpty` field check is set on test level, hence letting the tests pass. However, in the source code of

nico-kn:
As outlined in https://github.com/googleapis/nodejs-pubsub/issues/2117, the condition to enter the `WaitForProcessing` closing behavior of a subscription was checking against an `undefined` field `.isEmpty` of the inventory instead of checking for the `isEmpty()` method of the inventory.

In unit tests this went unnoticed since the field was set accordingly in the class under test.

This implicitly changes the default behavior of the subscription close functionality since the previous PR that introduced this functionality (https://github.com/googleapis/nodejs-pubsub/pull/2068) changed the default behavior without effect as the condition was never entered during `close()`.

___

- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/nodejs-pubsub/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass

Fixes #2117 🦕

### DIFF
--- a/src/lease-manager.ts
+++ b/src/lease-manager.ts
@@ -106,6 +106,8 @@ export class LeaseManager extends EventEmitter {
    * Adds a message to the inventory, kicking off the deadline extender if it
    * isn't already running.
    *
+   * @fires LeaseManager#full
+   *
    * @param {Message} message The message.
    * @private
    */
@@ -141,6 +143,10 @@ export class LeaseManager extends EventEmitter {
   }
   /**
    * Removes ALL messages from inventory, and returns the ones removed.
+   *
+   * @fires LeaseManager#free
+   * @fires LeaseManager#empty
+   *
    * @private
    */
   clear(): Message[] {
@@ -197,6 +203,7 @@ export class LeaseManager extends EventEmitter {
    * messages are left over.
    *
    * @fires LeaseManager#free
+   * @fires LeaseManager#empty
    *
    * @param {Message} message The message to remove.
    * @private
--- a/src/subscriber.ts
+++ b/src/subscriber.ts
@@ -1000,7 +1000,7 @@ export class Subscriber extends EventEmitter {
     const shutdownStart = Date.now();
     if (
       behavior === SubscriberCloseBehaviors.WaitForProcessing &&
-      !this._inventory.isEmpty
+      !this._inventory.isEmpty()
     ) {
       const waitTimeout = timeout.subtract(FINAL_NACK_TIMEOUT);
 
--- a/src/subscriber.ts
+++ b/src/subscriber.ts
@@ -655,7 +655,7 @@ export class Message implements tracing.MessageWithAttributes {
  *     close() function.
  * @property {SubscriberCloseBehavior} [options.behavior] The behavior of the close operation.
  *     - NackImmediately: Sends nacks for all messages held by the client library, and
- *       wait for them to send.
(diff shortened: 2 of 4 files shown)


nico-kn:
@feywind would be great if you could review this change. Linting and tests pass locally. 

However, the original `WaitForProcessing` feature was released as default behavior and this behavior change came with a minor version increment to 2.5.0. I would propose to release this as 3.0.0 or 2.6.0 instead of 2.5.1 since the default behavior change is a runtime relevant change for services consuming this client library.

feywind (dismissed):

feywind:
> However, the original `WaitForProcessing` feature was released as default behavior and this behavior change came with a minor version increment to 5.2.0. I would propose to release this as 6.0.0 or 5.3.0 instead of 5.3.1 since the default behavior change is a runtime relevant change for services consuming this client library.

I'm okay with it being a patch release since it's a bug in the intended behaviour. The fix would actually bring it closer to where it was before the shutdown options change.

I will double-check though.

kamalaboulhosn (dismissed):

sofisl (dismissed):

feywind:
I was hoping to avoid having to merge, but the test looks stuck.

feywind:
There was a legit bug in the system test that this exposed. Two, really, if you count the Duration change from before. Yay! Unfortunately I will have to see who's around for approvals.

sofisl (dismissed):

feywind (approved):

danieljbruce (approved):

PR #2122, opened 2026-01-12
tests: clean up system tests with separate resources per test

feywind:
System tests were potentially stepping on each other by sharing resources, as was exposed by [fixes to the new subscriber close method](https://github.com/googleapis/nodejs-pubsub/pull/2118). This should make them more deterministic.


system-test/pubsub.ts
```diff
@@ -417,352 +356,318 @@ describe('pubsub', () => {
...
     });
   });
 
   describe('Subscription', () => {
-    const TOPIC_NAME = generateTopicName();
-    const topic = pubsub.topic(TOPIC_NAME);
+    const thirtySecs = Duration.from({seconds: 30});
+    const sixtySecs = Duration.from({seconds: 60});
+    const thirty = {minAckDeadline: thirtySecs, maxAckDeadline: thirtySecs};
+    const sixty = {minAckDeadline: sixtySecs, maxAckDeadline: sixtySecs};
```
hongalex: I found this these consts unnecessary and made understanding the code hard to read. Are these values even important to the tests they're used in?

If not, it might be more clear to just inline them in the `subPop` function.

sofisl (approved):

PR #2127, opened 2026-02-10
fix: clean up earlier refactor issues with maxExtensionTime

### LINKED ISSUE #2054: 🐞 Bug: maxExtensionTime Ignored in Subscription Constructor

TL;DR: When initializing a `Subscription` with a custom `maxExtensionTime`, the value is silently ignored and the default one (1h) is used instead.

Longer Story:
In [this PR](https://github.com/googleapis/nodejs-pubsub/pull/2024), `maxExtensionMinutes` was refactored to `Subscriber.maxExtensionTime`. However, the logic that handled custom values via `setOptions` was not carried over.

As a result, subscriber always uses the default `maxExtensionTime` from the constructor. When a custom `maxExtensionTime` is passed, it’s not applied, because [Subscriber.setOptions](https://github.com/googleapis/nodejs-pubsub/blob/9ad808adb99b6e64c2c4431991721deae14b6400/src/subscriber.ts#L963) doesn’t override the default.

https://gist.github.com/AlessandroSteri/67f0139cff1b92a5d31e47b5d99df99f

1. Create a pubsub `Subscription` passing `maxExtensionTime` different from the default 1h as option
2. Observe that `subscription._subscriber.maxExtensionTime` is 1h and not the custom one

Bug: When initializing a Subscription with a custom `maxExtensionTime`, the value is silently ignored and the default one (1h) is used instead.

Expectation: `Subscriber.setOptions` should set the custom `maxExtensionTime`

- It's a bug introduced in the aforementioned refactor
- It's clear that the behaviour of that option is to configure the underlining subscriber

feywind:
The `maxExtensionTime` field was not being carried over into `Subscriber`, from a refactor a while back of various settings. This fixes that and adds a unit test for it.

This also patches out an old/faulty bit of documentation about `maxExtensionMinutes`.

Fixes https://github.com/googleapis/nodejs-pubsub/issues/2054 🦕

### DIFF
--- a/src/lease-manager.ts
+++ b/src/lease-manager.ts
@@ -51,8 +51,6 @@ export interface FlowControlOptions {
  * @property {number} [maxBytes=104857600] The desired amount of memory to
  *     allow message data to consume. (Default: 100MB) It's possible that this
  *     value will be exceeded, since messages are received in batches.
- * @property {number} [maxExtensionMinutes=60] The maximum duration (in minutes)
- *     to extend the message deadline before redelivering.
  * @property {number} [maxMessages=1000] The desired number of messages to allow
  *     in memory before pausing the message stream. Unless allowExcessMessages
  *     is set to false, it is very likely that this value will be exceeded since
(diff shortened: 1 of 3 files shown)


quirogas (approved):
LGTM 🚀

